PR #32, opened 2016-06-28
#WAVE2-324 Added condition for offset contextual link to only apply after a hero-cta class

dpfitzsi:
This should make the CSS only apply when there is a large block that would require the margin of the first contextual link to be forced down.


dpfitzsi:
Seems to be conflicting with some more recent commits. Need to close this request and put in a new version.

PR #50, opened 2016-07-06
Hotfix for faculties pages - ucgy-gbl-drawer.tpl

balrajb:


balrajb:
Changes been commited directly to develop.

PR #53, opened 2016-07-06
#WAVE2-324 Fixed Contextual links top margin

dpfitzsi:
Now a straight-forward conditional statement that finds whether a sibling of the first contextual-links-wrapper is a hero-cta and then sets the margin.


dpfitzsi:
Wrong commit slipped in again.

PR #76, opened 2016-07-12
#376 - Removed one extra div from main tpl file that caused big problems

robbdavis:


gsowrey:
This has already been addressed; branch no longer required.

PR #99, opened 2016-07-21
Removing the extra fields from banner brick

nindusaini:


dpfitzsi:
Hi Navinder, it looks like this pull request has some conflicts that need to be resolved before it can be merged in. Can you re-sync your branch with develop, make your changes and then resubmit?

dpfitzsi:
Navinder is making one change as result of the another merged pull overwriting some values.

PR #105, opened 2016-07-26
updating the webmail link from 365 to Webmail system

nindusaini:


dpfitzsi:
Thanks Navinder. Can this be changed from http://webmail.ucalgary.ca/ to https://webmail.ucalgary.ca. The http needs to redirect to https anyway so it would be best for traffic to get there directly.

nindusaini:
Updated the link to https

PR #106, opened 2016-07-27
Wave2 397 content ownership entities

balrajb:


dpfitzsi:
Hi Balraj, I have taken a look over this and 398 and 399 and it seems the admin.inc and .module have quite a bit of commented out code. Will this code be used in future dev or was it for debug? If it's the latter it should be removed before merge.

PR #109, opened 2016-07-27
Remove CSS hyphens support

TDRoden:
This was causing hyphens to show up all over the place in Safari and
Firefox.


dpfitzsi:
Hi Terry, it looks like this request has some conflicts. Can this be re-synced with develop, the changes reapplied and resubmitted? Thanks!

PR #119, opened 2016-08-02
394 - remove breadcrumb bar when breadcrumb disapears on small screens

robbdavis:
Campaign pages - remove breadcrumb bar on mobile.


dpfitzsi:
No longer a valid entry in Jira, Robb has confirmed this is no longer needed.

PR #124, opened 2016-08-05
Wave2 399 content ownership mini panels

balrajb:


gsowrey:
Balraj, can you add some comments to the code? Just so the next poor shmuck along the way doesn't have their brain ooze out through their ears trying to understand it. ;)

balrajb:
@gsowrey sure

gsowrey:
Could you throw some more into ucgy_gbl_content_ownership/ucgy_gbl_content_ownership.module? That's a pretty core one. 

Also, as a fundamental question (as I'm looking at this): Shouldn't this be under ucgy_core, not ucgy_gbl? Yes, it's a "global" thing, but core was meant for the deep-level, must-have stuff.

balrajb:
It was bit hard to decide initially that it has to go in "gbl" or "core", initially both looked relevant to me, I just choose one :) however its easy to change, would like to change in same Pull request ?

gsowrey:
Yeah, please move. I'd really like to make sure that deep-level stuff is in ucgy_core. This will make sure we can properly filter out non-gbl stuff in the modules list.

balrajb:
Ok discussed with Geoff, offline/online :) 

The final location of this module we decided:
/sites/all/modules/ucalgary/

so it will look like:
/sites/all/modules/ucalgary/content_ownership

PR #127, opened 2016-08-05
#WAVE2-441 Adding the code for eyes high menu

nindusaini:


gsowrey:
Code is being merged into wave2-483 (which is managing wave2-512).

PR #135, opened 2016-08-09
Bug wave247 views cache was breaking view

robbdavis:
Forced caching broke the views sorting functionality.
Removing student exp view from the forced caching method.


gsowrey:
Could you update that method (I won't ask for the entire file...) with a comment or two? Just so it's clear what's going on.

robbdavis:
No problem.

PR #163, opened 2016-08-22
Story wave2 464 eyes high italicising

robbdavis:
wave2-446 Eyes High Italicizing


gsowrey:
I think this is a decent MVP approach that allows us to make improvements in the future. The big thing right now is making sure you don't get multiple em's during the loop. I think you need to add in an if prior to the preg_match (eg on line 21, which is currently blank), like this:

if (strpos($text,'<em>Eyes High</em>') === false) {
    preg_match...
}

That way it won't run more than once. We can genericize that later on so we can use any string/tag combos without difficulty. But for now, it'll do.

Also: ADD COMMENTS! ;)

sites/all/modules/ucalgary/ucgy_core/ucgy_core_filters/ucgy_core_filters.module
```diff
@@ -0,0 +1,51 @@
...
+                // for each match add <em> tag
+                $text = str_replace($match, '<em>' . $match . '</em>', $text);
+            }
+    }
+  }
+  return $text;
+}
+
+/**
+ * @param $panels_pane
```
balrajb: Add 'Implements hook_fieldable_panels_pane_view().'

sites/all/modules/ucalgary/ucgy_core/ucgy_core_filters/ucgy_core_filters.module
```diff
@@ -0,0 +1,51 @@
...
+ * Loads the panel panes so we can get the title value of each pane.
+ * Does a search string replace on the title field.
+ */
+function ucgy_core_filters_fieldable_panels_pane_view($panels_pane) {
```
balrajb: Should be function ucgy_core_filters_fieldable_panels_pane_view($panels_pane, $view_mode, $langcode) {

dpfitzsi:
Robb, once you have the improvements mentioned by Geoff and Balraj this should be ready to merge once our cycle permits it.

robbdavis:
Thanks guys, working on this now.

robbdavis:
I'm a bit stumped on situations where a content editor italicized some Eyes High but not others in the same field. This renders Geoff's suggestion of: if (strpos($text,'<em>Eyes High</em>') === false) { unusable because we've got both.

Anyone know a regex that will only return Eyes High if it's not surrounded with <em>? Or, any other suggestions for the use case above.

gsowrey:
Sorry, Robb, I didn't get back to you sooner on this one. I think we need to change the preg_match_all('/\b(Eyes High)\b/', $text, $matches) to something else...

function _ucgy_core_filters_process($text) {
  $pattern = '/(Eyes High)(?![^<]*>|[^<>]*<\/)/gm';
  $new_text = preg_replace($pattern, "<em>Eyes High</em>", $text);
  // some sanity check maybe? 
  return $new_text;
}

robbdavis:
I've just uploaded a version using: '/(?!<em[^>]*?>)(Eyes High)(?![^<]*?</em>)/'.  Seems to work.

I've also included button text for str_replace.

robbdavis:
Last commit adds Eyes High italicising to mini-cta button text and breadcrumb. <em> tags are being added to menu but html is not rendering. need some help with that one.

PR #165, opened 2016-08-23
Wave2 424 requirement widget v2 comparision page

nindusaini:


balrajb:
- Not sure its an existing issue in Terry's code when, changing from Standard to Transfer, and click anywhere on page, the value did not get change. Same for Education System. Could be related to getting value from session.
- Code commenting please.
- Just usability question, can we use same code for requirements detail page for Filters?

balrajb:
one issue future-students/undergraduate/program-compare?compare=150

Educational system is not showing any info 
https://www.evernote.com/l/ALcRlVJl-JZOh7wAOjaxL6f3YFnJxUEofJ4

PR #168, opened 2016-08-23
WAVE2-32-adding link if no mini average is given

nindusaini:


balrajb:
I would recommend to put both text and link variables, in case we have change it.

Like 
`print l(variable_get('requirements_widget_no_min_ave_text','Please contact the admissions office.'), variable_get('requirements_widget_no_min_ave_link', 'https://choose.ucalgary.ca/request/requestInfo.do'); ?>
`

PR #171, opened 2016-08-24
WAVE2-494-correcting the target attribute for hero CTA

nindusaini:


robbdavis:
It might be just me but it doesn't seem to be working. For the top / main impact story, I selected 'open link in new window' option. When I then inspected the code on the home page I get this:
```
<a class="btn btn-ghost btn-lg" href="http://www.ucalgary.ca/utoday/issue/2016-08-08/ucalgary-grads-wrestle-olympic-gold-rio" role="button" target="">
                Read More              </a>
```

Target should = "_blank"

gsowrey:
While you're in that file, Navinder, can you add some PHP comments so we know what the various IF statements are doing, etc? Please and thank you! :)

PR #175, opened 2016-08-26
main branch for Requirement widget v2

balrajb:


balrajb:
Thanks for review @gsowrey few of improvements been done in other branch, could not merge into yet as we need stable probo for early deskcheck. 
However regarding multiple versions, 
- we had few stories originally created for V1 and higher priority, and V2 was not ready, developer (Navinder in this case) had to wait for V2 to complete, which could cause delays. So to speed up development, multiple versions helped a lot.
- Debuging and testing would be lot easier, can easily compare the behaviours for now, as there are number of scenarios.
So the plan is to remove V1 code as soon as V2 is stable.

gsowrey:
So long as the goal is to pull V1, that would be grand.

PR #181, opened 2016-08-26
WAVE2-525-making filters work on program details page

nindusaini:


balrajb:
some code commenting might help, but we still working on detail page can be done in next PR.

PR #182, opened 2016-08-29
433 using ['HTTP_HOST'] instead of  ['SERVER_NAME'] to detect root url

robbdavis:
433 pass protect live using [‘HTTP_HOST] instead of [SERVER_NAME’]. Test on http://solr-test3-ucalgary.pantheonsite.io/


dpfitzsi:
Robb confirmed this can be killed off.

PR #202, opened 2016-09-02
WAVE2-64-adding new fields for course translation text

nindusaini:


balrajb:
@nindusaini better to not show all province translation by default, let editors add more when they need. Should be a quick change.

balrajb:
thanks for changes Navinder

PR #213, opened 2016-09-13
437 Adding customized protected_pages module

robbdavis:
#437 Password protect specific pages or groups of pages.


gsowrey:
Ummmm... Why aren't we using the module in its original state? Why rename it all to UCGY? Given that our code is public, the original author might not be too pleased to see that... (Also, negates potential updates as it's disconnected from the drupal.org update path.)

gsowrey:
Right, I forgot that. 

Let's rewrite the README.txt file. Keep the accreditation for the work, but ditch everything else. Instead, put in notes that this is derived from the Drupal.Org project (provide URL and version), and then list the patch, and why it's there. Just so that the next person looking at this understands why.

robbdavis:
Will do.

robbdavis:
This got merged before I could fix up the readme. It's a minor change not affecting code.

I'll create a ticket and add this change to sprint 18 V11.

gsowrey:
*forehead slap*

I got ahead of myself. Sorry, Robb! I've restored the branch. Please update and I'll recommit.

robbdavis:
Hmm... I've made the changes/committ but Syncing / pushing doesn't seem to be doing anything.

robbdavis:
Created a new pull request here: https://github.com/ucalgary/www.ucalgary/pull/214

PR #244, opened 2016-09-23
Wave2 580 fpp list display

nindusaini:


gsowrey:
See sites/all/modules/contrib/cors/cors.module for an example of using hook_menu and building a custom admin form.

PR #254, opened 2016-09-29
569 Removing legacy homepage content type from ucgy_hp feature

robbdavis:
569 Removing legacy homepage content type from ucgy_hp feature.


robbdavis:
I removed ucyg_hp's dependency on the content type so that it could be deleted. Then figured deleting the content type manually would be cleanest.

Should I write the deletion into an update hook in ucgy_hp instead?

gsowrey:
If we don't need ucgy_hp anymore, then we should be making sure that it's being discontinued entirely, lock, shlock, and barrel. 

If necessary, let's do this in two parts: Disable first, then (in a second ticket) deletion.

robbdavis:
This story was just to delete the content type. We still need the ucgy_hp module. There are definitely sub modules in there (news and events widget)  that are needed. It looks like the templates in there are still needed too. I can create another story to clean up ucgy_hp now that the content type parts aren't needed.

gsowrey:
Ah, right, I forgot about those. 

Do we need the Strongarm settings? If we're yanking this, seems a bit odd that it's _adding_ something.

robbdavis:
Good call on removing the strongarm variables. I also removed content type fields.

But... that was a quick merge, I was going to ask you to do a quick glance over to make sure I haven't removed anything I shouldn't.

Should be all good though... site looks fine on new probo build.

dpfitzsi:
It looks like all that is in the strongarm file now is the xmlsitemap stuff. I assume that this module was probably added after the ucgy_hp content type was first created so it is now trying to add itself into the export. Since there is still a bunch of field_base and field_instance information in the ucgy_hp related to the content type this information makes sense to stay. A full delete of all field data (at least according to features) is not included in this diff. Features also won't delete files even if they have no relevance anymore so there may have to be some manual work here should all of this data be removed.

PR #272, opened 2016-10-10
419 enhancement views ownership

gsowrey:
A first crack at the Views ownership. It now needs a sanity check from someone else.


balrajb:
Just had a very quick look at files changed, first impression, do we need to update 4,833 files in this pull request? Can we update libraries/modules in separate pull request to make sure we are only focused on view content ownership in this PR? Will continue review...

gsowrey:
OH DAGNABBIT. I thought I’d committed that to a separate branch. That’s my fault.

gsowrey:
Er… For the function name? Not sure where you think the typo is…

balrajb:
Great, thanks Geoff for fix.

I am not sure this PR is suppose to complete the expected workflow or just code need review, I just tested the functionality in this probo:

1. Enable content ownership module
2. Add couple of taxonomy terms for ownership groups
3. Assign an ownership group to a test user
4. Assign same group to one of views
5. Login as that test user
6. Still don't have access to that view

gsowrey:
Hmm. That’s a bug. It should work all the way through at this point. If you log in as the admin and go to the views list, can you see the correct group assigned?

balrajb:
Yes I can see groups assigned to that view https://www.evernote.com/l/ALcA06roc3xPB448RHzT81HjMdtL3Cr8ZCQ

gsowrey:
Okay, I think I've figured out what's going on. 

Right now, a Content Admin doesn't have the permission to edit views. So we need that to be set in the Permissions tab in order to test properly. THEN this should all work. :)

balrajb:
ok, so we updating the permissions in separate PR? in that is a case, its ok to get this approved by QA.

PR #273, opened 2016-10-11
Story WAVE2 544 URL management

robbdavis:
WAVE2 544 URL management


sites/all/modules/ucalgary/ucgy_core/ucgy_core.module
```diff
@@ -460,3 +460,580 @@ function _ucgy_core_fpp_entity_clone($fieldable_panels_pane)
...
+    // If the user is a Site Admin or a Developer they can start the url path at 'root'
+    // All others see paths starting at a predefined page (like 'campaign' or 'future students')
+    if (in_array('Side Adminstrator', $user->roles) || in_array('Developer', $user->roles)) {
+        $paths = array("<root>");
+    } else {
+        $paths = array("Select");
+    }
+
+    // this taps into hook menu and passes the variables via drupal_alter
+    $menu_items = module_invoke_all('menu');
```
gsowrey: So invoke_all isn't going to cut it. We need to deal with the following tables: 

- menu_router
- page_manager_pages (notably the 'path' field; not sure the 'menu' field needs updating)

Instead, I think we need to consider using something like this: 

$menu_links = array();
foreach (menu_get_menus(false) as $menu) { // menu_get_menus(false) should exclude system menus
  $menu_links[$menu->name] = menu_tree_all_data($menu->name);
}

When we do an update, we should consider looking at the $menu_links array as the source. If there's a change to a specific $menu_links[$menu->name][$item], we:
- update the hierarchy within the $menu_links array
- update the menu link, using menu_link_save($new_item, $menu_links[$menu->name][$item])
- Find the appropriate Page Manager page, and update its path (there might be an API approach for this, but we can go to page_manager_pages if we have to)

And when it's all done, we should also run menu_rebuild, for sanity.

sites/all/modules/ucalgary/ucgy_core/ucgy_core.module
```diff
@@ -460,3 +460,580 @@ function _ucgy_core_fpp_entity_clone($fieldable_panels_pane)
...
+
+                // Creates an array of the url sections from the paths we have pulled from the db
+                $url_sections = explode('/', $only_wanted_paths);
+
+                // We've only update parent paths that are
+                // If the url section being changed is the same as the active leaf, we update the db
+                if ($url_sections[$count] == $old_active_leaf) {
+
+                    // This updates all the paths we want to change with the new path
+                    db_update('page_manager_pages')
```
gsowrey: Let's leave this as a last resort, too. Ideally, there's an API that'll handle this for us.

sites/all/modules/ucalgary/ucgy_core/ucgy_core.module
```diff
@@ -460,3 +460,580 @@ function _ucgy_core_fpp_entity_clone($fieldable_panels_pane)
...
+/**
+ * Helper function that returns list of menu items
+ */
+function _ucgy_return_menu_items(){
+    // We define $user so we can check roles
+    global $user;
+
+    // If the user is a Site Admin or a Developer they can start the url path at 'root'
+    // All others see paths starting at a predefined page (like 'campaign' or 'future students')
+    if (in_array('Side Adminstrator', $user->roles) || in_array('Developer', $user->roles)) {
```
robbdavis: Doh!

sites/all/modules/ucalgary/ucgy_core/ucgy_core.module
robbdavis: OK

I'll look into this

sites/all/modules/ucalgary/ucgy_core/ucgy_core.module
robbdavis: OK I'll look into the api.... may need to touch base with you on this later.

dpfitzsi:
Would need to be redone for future page manager work so this will be closed.

PR #288, opened 2016-10-19
Imp wave2 575 full width brick conversion

TDRoden:
Made sure all bricks were wrapped as content-fluid. This change ended up touching most bricks as some additional related class clean-up was performed.


dpfitzsi:
Hi Terry, it looks like we have a conflict with sites/all/modules/ucalgary/ucgy_gbl/ucgy_gbl_text_brick/templates/fieldable-panels-pane--ucgy_gbl_text_brick_fpp.tpl.php. Can you resolve it on your end?

TDRoden:
*whew* That ended up being far more than ucgy_gbl_text_brick_fpp.tpl.php!

PR #294, opened 2016-10-24
WAVE2-473- special bg video error improvements

nindusaini:


nindusaini:
I had those fields required but it was not throwing any errors for some reason, hence the custom solution. But I featurized this change as you suggested to be sure.

PR #322, opened 2016-11-08
Imp wave2 428 masquerade css improvements

TDRoden:
Block should be enabled by adding to the ‘Content’ region.


dpfitzsi:
Hi Terry, we've got some conflicts now with the drupal.css file and drupal.css.map files.

TDRoden:
I think it should be good now Dan - thanks!

PR #325, opened 2016-11-09
WAVE2-637 Added access arguments for FPP overview page to require admin access

dpfitzsi:


dpfitzsi:
It was decided that this will no longer be required because the new page manager being developed in 652 will make this dev work irrelevant.

PR #345, opened 2016-11-23
Bug wave2 741 rounded brick against rounded brick

TDRoden:


dpfitzsi:
Great work here Terry, but looks like we have a conflict in the bricks.css.map file before I can commit.

PR #346, opened 2016-11-23
WAVE2 655 Improve related content brick

robbdavis:
655 Removing hidden divs. Using only one (wide) image style for all chicklets.


dpfitzsi:
Thanks for this Robb and the comments about the commented code!

PR #349, opened 2016-11-24
Bug wave2 738 last brick rounded

TDRoden:


TDRoden:
Yeah, I just have to be aware of their presence when doing some of this spacing/padding logic!

PR #351, opened 2016-11-24
Wave2 560 merge test

gsowrey:


gsowrey:
This corrects issues we had merging 560 earlier.

PR #354, opened 2016-11-25
Imp wave2 671 image brick refinements

TDRoden:


dpfitzsi:
Hi Terry, just for clarification, the cleanup to the template files was mostly to remove the commenting about the wide variety of styles in the templates as these are now going to be defined by the CSS included in the brick? I have also seen that you have removed a couple of instances of inline styles (always a good thing).

TDRoden:
Yes, the cleanup was to remove unnecessary output - there were two output options dependent on location used but weren't necessary when the layout was properly Bootstrapped. Making every effort to remove HTML comments and inline styles too!

PR #359, opened 2016-11-28
WAVE2-758-adding new relic deploy marker

nindusaini:


nindusaini:
Yes Geoff, I have that file inside quicksilver folder. Is that not the right place???

gsowrey:
It needs to be in the site root, along with the .probo.yml file.


On Nov 29, 2016, at 10:52, Navinder <notifications@github.com<mailto:notifications@github.com>> wrote:


Yes Geoff, I have that file inside quicksilver folder. Is that not the right place???

-

nindusaini:
Ok, moved it to root folder.

pantheon.yml
```diff
@@ -0,0 +1,37 @@
...
+php_version: 7.0
+
+# Drush Version
+# https://pantheon.io/docs/pantheon-yml/#drush-version
+drush_version: 8
+
+# Protected Web Paths
+# https://pantheon.io/docs/pantheon-yml#protected-web-paths
+protected_web_paths:
+   - /example.txt
```
gsowrey: Let's make sure the example stuff is removed. ;)

pantheon.yml
```diff
@@ -0,0 +1,37 @@
...
+   - /example_directory
+
+# Nested Docroot
+# https://pantheon.io/docs/pantheon-yml#nested-docroot
+web_docroot: true
```
gsowrey: I think this needs to be false, we're not hosting a nested webroot. See https://pantheon.io/docs/pantheon-yml/

pantheon.yml
```diff
@@ -0,0 +1,37 @@
...
+# https://pantheon.io/docs/pantheon-yml#quicksilver
+workflows:
+  # Log to New Relic when deploying to test or live.
+    deploy:
+      after:
+        - type: webphp
+          description: Log to New Relic
+          script: private/quicksilver/new_relic_deploy.php
+    # Also log sync_code so you can track new code going into dev/multidev.
+    sync_code:
```
gsowrey: Do we need the sync_code?

private/quicksilver/new_relic_deploy/new_relic_deploy.php
```diff
@@ -0,0 +1,75 @@
...
+  // Get the annotation:
+  $changelog = `git tag -l -n99 $deploy_tag`;
+  $user = $_POST['user_email'];
+}
+
+
+// Use New Relic's v1 curl command-line example.
+// TODO: update to use v2 API with JSON, plus curl() in PHP.
+// Blocked by needing the app_id to use v2 API
+$curl = 'curl -H "x-api-key:'. $data['api_key'] .'"';
```
gsowrey: So what does the data in New Relic look like? Are we getting data for develop, dev, test, and live? Or only live?

pantheon.yml
nindusaini: Removed, but added pantheon.yml here, so its hidden from public accees

pantheon.yml
nindusaini: I thought we do want to monitor how code is sync'd, so we know where/if an issue starts.

pantheon.yml
nindusaini: Changed to false.

gsowrey:
Hmm. ‘Tis a good point. Yeah, maybe leave that on for now. We can always disable it if it doesn’t prove to be useful.

gsowrey:
What’s the data look like? Is it useful information in New Relic?

PR #372, opened 2016-12-04
WAVE2-560-revert-contrib-and-core-label-changes

dpfitzsi:
Fixes a few files that were merged and would be overridden in future
core/module updates.


dpfitzsi:
Will be reverted by 7.53 update so this is not needed.

PR #387, opened 2016-12-09
803 - Fixing merge conflicts in related content brick

robbdavis:
803 - Fixing merge conflicts in related content brick


robbdavis:
The reset script probably failed because the related content feature "needed review". I reverted it on the probo site and it seems to be working as it should.

PR #388, opened 2016-12-12
wave2-776 Hero - Large w/ Text Overlay - required fields, char count …

mhughes2012:
…and new fields


sites/all/modules/ucalgary/ucgy_core/ucgy_core_layout/ucgy_core_layout.module
```diff
@@ -595,3 +595,20 @@ function ucgy_core_layout_panels_pre_render($panels_display, $renderer)
...
+ * Implements hook_js_alter().
+ *
+ */
+function ucgy_core_layout_js_alter(&$javascript) {
+
+  // swap for a patched version of modules/contrib/maxlength/maxlength.js
+  $jsfile = drupal_get_path('module', 'maxlength') . '/js/maxlength.js';
+  foreach ($javascript as $path => &$values) {
+    if (is_string($values['data']) && strpos($values['data'], $jsfile) !== FALSE) {
+      $values['data'] = 'sites/all/themes/seven/js/maxlength.js';
```
dpfitzsi: Why is there a reference to the maxlength js in the seven theme? Shouldn't this js file be added by the line below in the ucgy_core_layout module?

modules/image/image.field.inc
```diff
@@ -422,6 +422,7 @@ function image_field_widget_process($element, &$form_state, $form) {
   $element['alt'] = array(
     '#title' => t('Alt text for images'),
     '#type' => 'textfield',
+    '#required' => 'true',
```
dpfitzsi: Minor change but this would need to be taken out as this is modifying core. 

This would be a better way of handling this kind of change. More code, but we can maintain it: https://evolvingweb.ca/blog/required-alt-text-images-drupal

mhughes2012: Thank you. This is a far superior solution. My apologies for adding this to core, I'm not sure why I did that.
Fixed now.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_layout/ucgy_core_layout.module
mhughes2012: I'm positive this has been removed in other PRs, but I removed it here too.

dpfitzsi:
Thanks for the updates Mike! Can you also update the ticket on the Jira end that this needs to be merged after your other field standardization tickets because this will correct the other instances of the field.image.inc file that is present in those ones.

dpfitzsi:
This one is also in conflict so this likely to be the case as we work through each of these tickets.

mhughes2012:
Hi @dpfitzsi, Merged.

dpfitzsi:
Merged in 419

PR #389, opened 2016-12-13
WAVE2-808 - content ownership fix 

balrajb:


dpfitzsi:
Looking pretty good in general, but there is a .gitignore file ignoring the default.settings.php file which has blown away that file as well. Can this be removed?

balrajb:
@nindusaini any reason for deleting default.settings file?

balrajb:
@dpfitzsi reverted and default setting back in here. thanks for review

PR #393, opened 2016-12-14
773 - Cloned bricks now inherit content ownership groups from original brick

robbdavis:
773 - Cloned bricks now inherit content ownership groups from original brick


robbdavis:
Hi Dan,

For the life of me, I couldn't find a more elegant way to get the PID of the page being cloned.
What I did was pull the page name from the current path, load the page using the page name and then used $page-pid when I inserted the groups into the content ownership table.

See lines 333 - 338 (and the db insert line 361).

This works but I'm wondering if I'm missing a simple way to do this without having to manipulate the string on line 334?

sites/all/modules/ucalgary/ucgy_core/ucgy_core.module
```diff
@@ -326,6 +326,50 @@ function ucgy_core_panels_display_save($display)
         //initialize $new_content variable as an array
         $new_content = array();
 
+        // Get the global $user
```
dpfitzsi: Can you change this to // Get the current user

sites/all/modules/ucalgary/ucgy_core/ucgy_core.module
```diff
@@ -326,6 +326,50 @@ function ucgy_core_panels_display_save($display)
...
+        global $user;
+
+        // This is a way we can get the page name
+        $url = current_path();
+        $page_name = substr($url, strlen('admin/structure/pages/edit/page-'));
```
dpfitzsi: I took a look at this one and there doesn't seem to be a "better" way from what I can see. This information is not stored in the panels_display table in the format you need it. It would be either you find a way to bypass the call on 338 and just pass the pid directly on 361 or go with this method.

sites/all/modules/ucalgary/ucgy_core/ucgy_core.module
```diff
@@ -326,6 +326,50 @@ function ucgy_core_panels_display_save($display)
...
+        // If there are results
+        if (!empty($results)) {
+
+            // Set a counter for the delta value
+            $i = 0;
+
+            // Loop through each result
+            foreach ($results as $result) {
+
+                // Create a new content ownership record for the page using the users groups
```
dpfitzsi: change users to user's

robbdavis:
Hi Dan I've added a new form alter function that only allows certain roles (not content editors) to clone.

dpfitzsi:
New form alter looks good. Will pass on to Louise. Thanks Robb!

dpfitzsi:
Hi Robb can you resolve the conflict with the ucgy_core.module?

robbdavis:
Hi Dan,

Conflicts resolved. I tested the resolved branch on my local and it's all working as expected.

PR #398, opened 2016-12-15
Subtask wave2 796 hero cta required field char counts new fields

mhughes2012:


themes/seven/style.css
```diff
@@ -1006,4 +1012,15 @@ div.add-or-remove-shortcuts {
   border-radius: 8px;
 }
 
+#field-uc-banner-bg-color-add-more-wrapper label:after {
```
dpfitzsi: Also include the styles you have entered here to your new seven_style.css file.

mhughes2012: This exists in the new seven_styles.css, added now in PR 406

PR #400, opened 2016-12-16
Wave2 732 merge this branch

nindusaini:


robbdavis:
Hi Navinder, 

There might be something I'm missing as to why you put them there, but it is generally bad Drupal practice to put sql queries (or any complex php functions for that matter) into .tpl files. See here:
[http://drupal.stackexchange.com/questions/41007/execution-of-complex-db-queries-in-tpl](url)

So the question is how do you pass the sql results to the theme? I think views_pre_render is the answer.

You could try this in the .module file:

`function content_ownership_bulk_views_pre_render(&$view) {

     if ($view->name == 'group_specific_content_fpp') {
         foreach ($view->result as $result) {
             $dbquery = db_select('content_ownership', 'co')
                 ->fields('co', array('entity_group'))
                 ->condition('co.entity_id', $result->fpid, '=')
                 ->condition('co.subtype', 'fieldable_panels_pane', '=')
                 ->condition('co.deleted', '0', '=')
                 ->execute()
                 ->fetchAll();

             $view->term = $dbquery;

         }
     }

}`

You would have to write a query for each view (i.e. one using nid, and one for pid).

Then in the template (e.g. views-view-field--group-specific-content-fpp--fpid-1.tpl):
`$term_info = taxonomy_term_load($view->term[0]->entity_group);
 print $term_info->name.'<br/>';`

The above seems to work for me but, again, I might be missing something. Also I may have screwed up how this loops through the content so definitely double check it.

Also, please confirm with @dpfitzsi if I'm correct about this. Your code does work, and it is possible yours is the correct way in this circumstance.

nindusaini:
Thank you for the code review Robb!! This is a much cleaner solution, implementing it.

sites/all/modules/ucalgary/content_ownership/content_ownership_bulk/templates/views-view-field--group-specific-content-fpp--fpid-1.tpl.php
```diff
@@ -0,0 +1,18 @@
...
+<?php
+$terms = $view->term[$output][0];
+//if we have more than three groups we'll display ... after the third one
+$count = count($view->term[$output][0]);
+if($count > 2){
+    for($i=0;$i<3;$i++){
+        $term_info = taxonomy_term_load($view->term[$output][0][$i]->entity_group);
+        print $term_info->name.'<br/>';
+    }
+    print '...';
```
dpfitzsi: Was this just test code?

sites/all/modules/ucalgary/content_ownership/content_ownership_bulk/templates/views-view-field--group-specific-content-nodes--nid.tpl.php
```diff
@@ -0,0 +1,18 @@
...
+<?php
+$terms = $view->term[$output][0];
+//if we have more than three groups we'll display ... after the third one
+$count = count($view->term[$output][0]);
+if($count > 2){
+    for($i=0;$i<3;$i++){
+        $term_info = taxonomy_term_load($view->term[$output][0][$i]->entity_group);
+        print $term_info->name.'<br/>';
+    }
+    print '...';
```
dpfitzsi: Also just checking if this is test code...

sites/all/modules/ucalgary/content_ownership/content_ownership_bulk/templates/views-view-field--group-specific-content-pages--pid-1.tpl.php
```diff
@@ -0,0 +1,19 @@
...
+<?php
+$terms = $view->term[$output][0];
+//if we have more than three groups we'll display ... after the third one
+$count = count($view->term[$output][0]);
+if($count > 2){
+    for($i=0;$i<3;$i++){
+        $term_info = taxonomy_term_load($view->term[$output][0][$i]->entity_group);
+        print $term_info->name.'<br/>';
+    }
+    print '...';
```
dpfitzsi: Also just checking if this is test code...

PR #402, opened 2016-12-16
Bug wave2 734 rounding background image bricks

TDRoden:


TDRoden:
I am going to close this pull request. I built it based off of WAVE2-699. With all of the problems around WAVE2-699 - and given that it is now merged - I would feel safer starting with a fresh branch.

PR #405, opened 2016-12-16
Subtask wave2 797 image required fields char counts new fields

mhughes2012:


mhughes2012:
@dpfitzsi the 'Caption for image'  does have character counts implemented, but I don't think there has been any change except for the maxlength js alter

PR #408, opened 2016-12-20
Wave2 758 new relic deployment marker

mhughes2012:
updated references to the new relic deploy script in pantheon.yml


dpfitzsi:
Hi Mike, I have tested this on newrelic multidev and can confirm it does work. Thanks for fixing the path!

PR #411, opened 2016-12-23
WAVE2-834 update probo.yml to remove rr and enable of content ownership

dpfitzsi:


balrajb:
Looks good, I think we should remove this too   
- drush rr

dpfitzsi:
Done, thanks Balraj!

PR #427, opened 2017-01-06
[wave2-796] remove rounded corners field from Hero CTA

mhughes2012:


sites/all/modules/ucalgary/ucgy_gbl/ucgy_gbl_cta/ucgy_gbl_cta.install
```diff
@@ -6,7 +6,7 @@
 
 
 //Deleting fields that are no longer needed.
-function ucgy_gbl_cta_update_7406()
+function ucgy_gbl_cta_update_7408()
```
dpfitzsi: I would just change this to 7001(). This would mean this is the first 7.x update for this module. This number means this is the 8th update for the 4.x branch of the 7.x core version of this module.

mhughes2012: @dpfitzsi, thanks for the info, I've updated it to 7001, although that didn't seem to be run. Investigating.

dpfitzsi:
Ah, I can see from the history of the file that Navinder originally started this with 7406, so yours should then be 7407. Since your correct number is earlier in the sequence it won't run. We can look into correcting the numbers in the future.

mhughes2012:
@dpfitzsi, as per our conversation, I've created a new update function with only the rounded corners field in there. Looking good now from my end.

PR #444, opened 2017-01-16
WAVE2-856-global drawer updates for fgs contact page

nindusaini:


sites/all/modules/ucalgary/ucgy_gbl/ucgy_gbl_drawer/ucgy_gbl_drawer.install
```diff
@@ -39,3 +39,16 @@ function ucgy_gbl_drawer_update_0002(){
     }
 
 }
+
+//Deleting fields that are no longer needed.
```
dpfitzsi: hook_update_N should be setup using the same block commenting that you would use in functions. The reason for this is that this information gets read is an is inputted during the update process as a description of what the update is doing.

sites/all/modules/ucalgary/ucgy_gbl/ucgy_gbl_drawer/ucgy_gbl_drawer.install
```diff
@@ -39,3 +39,16 @@ function ucgy_gbl_drawer_update_0002(){
...
+function ucgy_gbl_drawer_update_0003()
```
dpfitzsi: Change number to 7003, we should correct the version numbers to be based on 7.x branch

nindusaini: Hi Dan, corrected these two things, please if you can change the base branch to appropriate FGS before the merge (whenever that happens) would be great.

PR #453, opened 2017-01-22
Wave2 923 sso user lockout

gsowrey:
Build to test SSO codification and user lockout code


.probo.yml
```diff
@@ -45,6 +45,7 @@ steps:
       - drush en content_ownership_core -y
```
dpfitzsi: This probo.yml file will be outdated by the release of V18.0. There is an update going in under 834 that will remove the content ownership and drush rr calls to improve the performance of creating our probos. Once that gets committed we can pull in the latest develop to fix this issue.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_roles/ucgy_core_roles.module
```diff
@@ -349,14 +482,17 @@ function has_admin_privileges($account = null)
...
+    // If dealing with User 1, we'll assume all is possible
+    if ($account->uid === 1) {
+        return true;
+    }
+
     // Define Role admin powers (for simplified testing)
     $admin_roles = array(
         'administrator',
         'Developer',
-        'Content Administrator',
```
dpfitzsi: Removing this will break Mike's views ownership clone. He had to have this in the array so that Content Admins could clone.

dpfitzsi:
Looks like we have conflicts with probo.yml (now that 834 is merged in V18.0) and settings.php has a conflict too.

PR #469, opened 2017-01-27
WAVE2 877 FGS degree supervisors view filtering

robbdavis:
WAVE2 877 FGS degree supervisors view filtering


sites/all/modules/contrib/views_autocomplete_filters/js/views-autocomplete-filters-dependent.js
```diff
@@ -0,0 +1,78 @@
...
+          db.cache[searchString] = matches;
+          // Verify if these are still the matches the user wants to see.
+          if (db.searchString == searchString) {
+            db.owner.found(matches);
+          }
+          db.owner.setStatus('found');
+        }
+      },
+      error: function (xmlhttp) {
+        alert(Drupal.ajaxError(xmlhttp, db.uri));
```
mhughes2012: Let's log this error instead of as an alert to the browser

robbdavis: Hi Mike, I agree that errors in an alert box is a bad idea. Drupal 7 has this problem in a lot of places. That said, I don't think it is worth it to change this. It would mean patching the contrib module and that is something we want to avoid (and have miraculously avoided up to now).

robbdavis: A quick search did turn up this module. I could be something we enable if we think the issue is big enough:
https://www.drupal.org/project/prevent_js_alerts

mhughes2012: Agree that we don't want to hack a contrib module. This could be done with a hook js alter, but it might be overkill to accomplish this sort of nit pick.  Review approved with no changes.

sites/all/modules/ucalgary/ucgy_fgs/templates/views-view-fields--fgs-grad-degree-supervisors.tpl.php
```diff
@@ -0,0 +1,44 @@
...
+ *   - $field->label_html: The full HTML of the label to use including
+ *     configured element type.
+ * - $row: The raw result object from the query, with all data it fetched.
+ *
+ * @ingroup views_templates
+ */
+?>
+<!-- Faculty Info -->
+<div>
+        <?php  $url = strip_tags($fields['path']->content); ?>
```
dpfitzsi: I would just add a comment here to explain what is happening in this template. For example, you're grabbing the path (alias) of the GBL Supervisor content type and linking their names back to their supervisor page.

robbdavis: Hi Dan, makes sense. Can we make this change after QA /PO Desk check but before we merge? It would save me the 20 - 30 minutes of rebuilding everything and populating content for the view. Or should we just stick to the process?

dpfitzsi: Hi Robb, looks like probo has been reaped anyway. Can you make these commenting changes and then we can send back to Louise?

robbdavis:
HI Dan. Commenting added and merge conflicts fixed. This should be good to go to Louise now.

dpfitzsi:
Want to merge this to Develop, can you resolve these conflicts?

robbdavis:
Should be good now dan.

PR #470, opened 2017-01-30
WAVE2-936 Added check for empty columns in data table and remove if empty

dpfitzsi:


dpfitzsi:
Not needed as part of V18.0 release. It appears that I had a bad DB copy of some sort that was causing this issue. I will close this off and refer back if it is ever needed again.

Thanks Mike for the help in reviewing this PR.

PR #476, opened 2017-01-31
WAVE2-851-adding ref field for another banner

nindusaini:


nindusaini:
I'll open this PR again, need space for tickets in PO check

PR #482, opened 2017-02-01
Imp wave2 845 v2 grad details outcomes next steps

mhughes2012:
add removed isset for academic path for FGS Degree outcomes and next steps


sites/all/modules/ucalgary/ucgy_fgs/templates/node--fgs-graduate-degree.tpl.php
```diff
@@ -67,22 +67,24 @@
...
-        </div>
+        <?php if (isset($content['field_fgs_grad_degree_acdmc_path'])) { ?>
+            <div class="ac_path">
+              <p><?php echo t('With a ') . $grad_degree .  t(' degree in ') . $grad_program . ', '; ?>
+              <?php
+
+                $ap_copy = render($content['field_fgs_grad_degree_acdmc_path']);
+                $firstp = '<p>';
+
+                $pos = strpos($ap_copy, $firstp);
```
gsowrey: Why are we stripping out the paragraph tag from the Drupal render, and not just dropping it from line 72? Seems like unnecessary work. If we need to alter the output of the Drupal field, we should do so on the preprocess end, not in the template.

sites/all/modules/ucalgary/ucgy_fgs/ucgy_fgs.module
```diff
@@ -47,7 +47,7 @@ function ucgy_fgs_form_alter(&$form, &$form_state, $form_id) {
     $ac_path_title = $form['field_fgs_grad_degree_acdmc_path']['und']['#title'];
     $ac_path_desc = $form['field_fgs_grad_degree_acdmc_path']['und']['#description'];
 
-   $form['field_fgs_grad_degree_acdmc_path']['und']['#prefix'] = '<p><strong>' . $ac_path_title . '</strong><br/>' . $ac_path_desc . '</p>' ;
+    $form['field_fgs_grad_degree_acdmc_path']['und']['#prefix'] = '<p><strong>' . t($ac_path_title) . '</strong><br/>' . t($ac_path_desc) . '</p>' ;
```
gsowrey: Okay, wait, we're ADDING the paragraph tag here, just to strip it out in the template??

mhughes2012: This moves the description (help) text in the edit form to above the text field so the user know that there will be copy added before whatever they type in the text box.

sites/all/modules/ucalgary/ucgy_fgs/templates/node--fgs-graduate-degree.tpl.php
mhughes2012: the render of the field wraps the content in markup, which we want -- except for the opening paragraph tag because we are prefacing the the content with a sentence fragment based on the degree and the program.

gsowrey: Mike has explained this, and it now makes sense.

sites/all/modules/ucalgary/ucgy_fgs/ucgy_fgs.module
gsowrey: Red herring; Mike explained.

PR #485, opened 2017-02-03
855 Template fix for case where span was added even if credentials are not added.

robbdavis:
855 Supervisor Availability


robbdavis:
Dan, I named this branch incorrectly. **It should be wave2-855**. Also, I had to use develop as the base because fgs is a mess after the most recent commit. Let me know if there is anything I need to correct from my end before merge.

PR #489, opened 2017-02-03
Imp wave2 842 v4 merge fix

dpfitzsi:


dpfitzsi:
Mike, Robb, please verify this new version of 842 to make sure there are no obvious issues here. I have pulled out the merge conflicts from the prevoius branch and manually resolved conflicts in the ucgy_fgs.info, ucgy_fgs.module, and ucgy_fgs template file.

robbdavis:
Well this looks much, much better ;). There are three files (video.module, related_content.module, and drawer.module) that only have some spacing changes but I'm not sure why they were touched at all. The don't seem related to the story.

We should maybe go over what went wrong with Mikes commit/merge (669eb2a). We should also go over the process which allowed that to get merged to fgs. It couldn't have been tested because the code in that commit wouldn't even allow a probo to get built (it returned an error).

PR #499, opened 2017-02-07
WAVE2-952-adding mini-cta ref field

nindusaini:


sites/all/modules/ucalgary/ucgy_fgs/ucgy_fgs.features.field_base.inc
```diff
@@ -138,8 +138,8 @@ function ucgy_fgs_field_default_field_bases() {
     'module' => 'list',
     'settings' => array(
       'allowed_values' => array(
-        'thesis-based' => 'Thesis based',
-        'course-based' => 'Course based',
+        'Thesis based ' => ' Thesis based',
```
dpfitzsi: Recording here this is fixed in 912.

sites/all/modules/ucalgary/ucgy_fgs/templates/node--fgs-graduate-degree.tpl.php
```diff
@@ -257,5 +257,32 @@
 
 <div><?php print views_embed_view('fgs_grad_degree_supervisors', 'block'); ?></div>
 
-
+<!--program highlights section -->
```
gsowrey: Please make HTML comments in PHP.

sites/all/modules/ucalgary/ucgy_fgs/ucgy_fgs.features.field_base.inc
gsowrey: This might sound really dumb, but why are we inserting a space into select values? Also, as a general rule, select options should be key=>value, where the key is machine-safe (no spaces, lowercase letters, preferably unique).

sites/all/modules/ucalgary/ucgy_fgs/templates/node--fgs-graduate-degree.tpl.php
nindusaini: Comments changed to PHP.

sites/all/modules/ucalgary/ucgy_fgs/ucgy_fgs.features.field_base.inc
nindusaini: You're right Geoff, we noted that and because this has different values in different stories, so instead of going to each and every story it was decided that this field will get fixed in wave2-912 the integration story with the correct keys and values as you mentioned.

PR #506, opened 2017-02-08
Imp wave2 955 supervisor help text

mhughes2012:
includes merge of 926 to capture the new way we are deciding which degrees are supervised


robbdavis:
Approved but noting conflict.

mhughes2012:
removed that conflict.

PR #509, opened 2017-02-10
WAVE2-852-creating branch again because of so many conflicts

nindusaini:


sites/all/modules/ucalgary/ucgy_fgs/templates/node--fgs-graduate-degree.tpl.php
```diff
@@ -267,6 +267,111 @@
     ?>
 </div>
 
+<<<<<<< HEAD
```
balrajb: this should not be here!

nindusaini: Thank you Balraj.

PR #523, opened 2017-02-17
Imp wave2 876 supervisor details page

mhughes2012:


sites/all/modules/ucalgary/ucgy_core/ucgy_core_variable_table_form/ucgy_core_variable_table_form.module
```diff
@@ -0,0 +1,188 @@
...
+  // set a variable to pass to the JSs
+  drupal_add_js(array('vtownership' => array('form_ownership' => variable_get('variable_table_form_ownership'))), 'setting');
+
+
+  $current_user = user_load($user->uid);
+  $ownership_groups = $current_user->field_user_ownership_group[LANGUAGE_NONE];
+
+  $arr_user_ownership_group = [];
+  foreach ($ownership_groups as $val) {
+    array_push($arr_user_ownership_group, $val['target_id']);
```
robbdavis: I think  $arr_user_ownership_group[] = $val['target_id'] would add the values to the array without needing array_push. Or, is array_push the better way to go in this situation? Just curious.

Either way, I think this could use some commenting to explain what you're doing here.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_variable_table_form/ucgy_core_variable_table_form.module
```diff
@@ -0,0 +1,188 @@
...
+  content_ownership_add_form_group_ownership_options($form, $form_state, 'node', $form['type']['#value'], $form['nid']['#value']);
+
+  $form['ownership']['#title'] = "Ownership Groups for this Page";
+  $form['ownership']['#access'] = $user->uid == 1 || in_array('Site Administrator', $user->roles);
+
+  return $form;
+}
+
+function ucgy_core_variable_table_form_form_validate($form, &$form_state) {
+}
```
robbdavis: This function doesn't seem to be doing anything. If it does need to be here for some reason, it needs a comment.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_variable_table_form/ucgy_core_variable_table_form.module
```diff
@@ -0,0 +1,188 @@
...
+
+function ucgy_core_variable_table_form_form_submit($form, &$form_state) {
+
```
robbdavis: I think it's good practice to have a comment for each function.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_variable_table_form/ucgy_core_variable_table_form.module
```diff
@@ -0,0 +1,188 @@
...
+    $ownership_results .= $val.",\n";
+
+  }
+
+  if (!valid_email_address($form_state['values']['fgs_email'])) {
+    if (empty($form_state['values']['fgs_email'])) {
+      form_set_error('submitted][email_address', t('The FGS Email address is required.'));
+    } else
+    form_set_error('submitted][email_address', t('The FGS Email address appears to be invalid.'));
+  }
```
robbdavis: This if statement should be explained in a comment (even if it's obvious to us what's happening).

sites/all/modules/ucalgary/ucgy_core/ucgy_core_variable_table_form/ucgy_core_variable_table_form.module
```diff
@@ -0,0 +1,188 @@
...
+ */
+function ucgy_core_variable_table_form_form_alter(&$form, &$form_state, $form_id) {
+
+  // Check for the supervisor node form
+  if ($form_id == 'ucgy_core_variable_table_form_form') {
+
+    //  Hide the fgs related fields if ucgy_fgs module  is not enabled.
+    if (!module_exists('ucgy_fgs')) {
+      hide($form['gblsupervisorformset']);
+    }
```
robbdavis: Just curious here. What if we want to add another formset on a site where the fgs module exists? For example what if we want to add some variables for 'undergraduate' ? Won't the validation and required fields conflict for each formset?  Would it make sense to use completely separate forms for each tab? I don't know maybe that's a tonne more work?

sites/all/modules/ucalgary/ucgy_core/ucgy_core_variable_table_form/ucgy_core_variable_table_form.module
```diff
@@ -0,0 +1,188 @@
...
+
+
+    drupal_add_js(drupal_get_path('module', 'ucgy_core_variable_table_form')  . '/js/ucgy_core_variable_table.js');
```
robbdavis: Comment for this function. Also, I've been told this style throws an error in php 5.6 and less but I might be wrong about that; still I think this is easier to read:
$path = drupal_get_path('module', 'ucgy_core_variable_table_form');
drupal_add_js($path . '/js/ucgy_core_variable_table.js');

mhughes2012: commented. I changed it to your style as well, that is more readable.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_variable_table_form/ucgy_core_variable_table_form.module
mhughes2012: The idea of this form is to be global, so what might have to happen is we check for module_exists before creating the form elements. That way we won't be validating items that we don't need to. I'll let @dpfitzsi weigh in here.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_variable_table_form/ucgy_core_variable_table_form.module
mhughes2012: commented.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_variable_table_form/ucgy_core_variable_table_form.module
mhughes2012: agreed. added the function comments.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_variable_table_form/ucgy_core_variable_table_form.module
mhughes2012: removed.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_variable_table_form/ucgy_core_variable_table_form.module
mhughes2012: I had more success with array_push in this situation, but I'm sure either method would work. I commented this section to explain what's happening.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_variable_table_form/ucgy_core_variable_table_form.module
robbdavis: Yeah... I see this variables table being pretty useful globally. In the case of ,say, an undergraduate or 'eyes high' section needing variables added, the fgs module will still exist so that check wouldn't work. It wasn't part of the scope /AC of this story (so should probably  be new story?) but I think each section should be it's own separate form. That would avoid any validation problems. @dpfitzsi  thoughts?

robbdavis:
Approving  but I do have questions around using just one form for different sections of the the site that require variables being added. See comments below line 179.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_variable_table_form/js/ucgy_core_variable_table.js
```diff
@@ -0,0 +1,17 @@
...
+
+    // get the variable table ownership for the form
+    var table_ownership = Drupal.settings.vtownership.form_ownership;
+
+    // loop through each ownership checkbox
+    $( "#edit-ownership .form-type-checkbox input" ).each(function() {
+
+      // if the table ownership string contains the value of this checkbox, check it
+      if (table_ownership.includes($(this).val())) {
+        $(this).attr("checked", "checked");
```
dpfitzsi: This should be .prop as of jQuery 1.6.

mhughes2012: thanks @dpfitzsi, this has been fixed.

sites/all/modules/ucalgary/ucgy_gbl/ucgy_gbl_supervisor/ucgy_gbl_supervisor.module
```diff
@@ -185,3 +186,37 @@ function ucgy_gbl_supervisor_default_rules_configuration() {
...
+  $replacements = array();
+
+  if ($type == 'node' && !empty($data['node'])) {
+
+    foreach ($tokens as $name => $original) {
+      switch ($name) {
+        // set the token value
+        case 'gbl-supervisor-change-email':
+          $replacements[$original] =  variable_get('fgs_email', 'jaela.curtis@ucalgary.ca');
+          break;
```
robbdavis: This is fine to leave this way but, generally, it is good practice to only use switch statements if you have a long list of cases. Otherwise a good old if statement is the way to go

sites/all/modules/ucalgary/ucgy_gbl/ucgy_gbl_supervisor/ucgy_gbl_supervisor.module
```diff
@@ -185,3 +186,37 @@ function ucgy_gbl_supervisor_default_rules_configuration() {
...
+{
+  $replacements = array();
+
+  if ($type == 'node' && !empty($data['node'])) {
+
+    foreach ($tokens as $name => $original) {
+      switch ($name) {
+        // set the token value
+        case 'gbl-supervisor-change-email':
+          $replacements[$original] =  variable_get('fgs_email', 'jaela.curtis@ucalgary.ca');
```
robbdavis: Should jaela's email be hardcoded?

mhughes2012: yes Jaela's is the default in the case that fgs_email is not set. (There is no reason for this not to be set, btw)

sites/all/modules/ucalgary/ucgy_gbl/ucgy_gbl_supervisor/ucgy_gbl_supervisor.module
mhughes2012: sorry, you are right of course. I was setting it up for more tokens, but that's overkill at this point.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_variable_table_form/ucgy_core_variable_table_form.module
```diff
@@ -265,10 +265,27 @@ function ucgy_core_variable_table_form_form_submit($form, &$form_state) {
   if (!valid_email_address($form_state['values']['fgs_email'])) {
     // if the email address is empty throw a required error
     if (empty($form_state['values']['fgs_email'])) {
-      form_set_error('submitted][email_address', t('The FGS Email address is required.'));
-    // otherwise throw an invalid error
-    } else
-      form_set_error('submitted][email_address', t('The FGS Email address appears to be invalid.'));
+      form_set_error('submitted][email_address', t('The Content Change Notification Email address is required.'));
+    }  else {
```
dpfitzsi: Extra space between bracket and else

PR #541, opened 2017-02-28
Wave2 1005 supervisor update template

mhughes2012:


sites/all/modules/ucalgary/ucgy_gbl/ucgy_gbl_supervisor/templates/node--gbl-supervisor.tpl.php
```diff
@@ -75,7 +75,7 @@
                             } else {
 
                                 // Print the name without any link
-                                print '<h1>'.$fname.' '.$lname.'<span class ="credentials">'.$abbreviated_credentials.'</span></h1>';
```
dpfitzsi: Can you remove the space between <span class ="credentials"> so that it reads <span class="credentials">?

mhughes2012: that line was deleted anyways. No changes made.

PR #543, opened 2017-03-02
Wave2 979 new page manager menu

nindusaini:


nindusaini:
Thanks for your comments Balraj,
Here's the explanation: 1. I think we won't select any menu by default, we'll just display existing menus and when user clicks on them then menu items will display, if you can please make a selection and then click on "add item" button and then hit "create menu item", then you won't see that error message.
2. is_avaiable function is what checks for duplication of existing urls 
3. AC says content admins but I will double check with Jaela.

balrajb:
@nindusaini thanks for fixes and explanation.

Looks good, just having some help notices/messages would help e.g. "please add item", "menu item <MENU ITEM NAME> created successfully" and "menu item can't be created because it is duplicate".

Also "add item" should be "Add menu item".

sites/all/modules/ucalgary/ucgy_core/ucgy_page_manager/ucgy_page_manager_menu_item.inc
```diff
@@ -0,0 +1,280 @@
...
+    $menu_name .= '</table>';
+    $menu_name .= '</td>';
+    $menu_name .= '</tr></table>';
+    return ($menu_name);
+}
+
+/*
+ * Helper function to get the menu_items according to their depth
+ */
+function get_childs($parent) {
```
balrajb: sorry for more comments but when you work on help text (user messages), can you also change "get_childs" to "get_children"

taeram:
Closing this PR on GitHub, but keeping the branch around for historical purposes.

PR #548, opened 2017-03-05
1002 - add js to hide filter when no values. add css to show only selected filter values. change filter name and order. remove fitler css from .module

robbdavis:
1002 - show auto selected filters on degree selection popup.


dpfitzsi:
Robb can you resolve these conflicts?

robbdavis:
This should be good now Dan! *hopes he didn't inadvertently blow away css.

PR #551, opened 2017-03-07
[wave2-1008] supervisor login view implementation

mhughes2012:


robbdavis:
The view works as expected. That said, I'm unsure about adding the view through the block system?@mhughes2012 can you confer with @dpfitzsi on this? I don't think we are really using our block system much at all and especially not to attach views. Maybe this is a special case, but I think the better way would be to print the view out programmatically on the user-page template. 

I am not sure if this view is for supervisors only but you could put in an if statement to only print the view for supervisors.

robbdavis:
note: also conflicts need to be resolved.

mhughes2012:
yes, I didn't like the block system being used here either. @dpfitzsi let me know if we want to change this. resolved conflicts.

dpfitzsi:
Using the block should be fine considering our timeline. This is something we may want to revisit anyway since the display is a bit off.

PR #552, opened 2017-03-08
Added new module for converting entity titles to ids

dpfitzsi:


dpfitzsi:
Not tied to a JIRA ticket because this is a one time import that will be deleted once completed.

PR #555, opened 2017-03-09
Wave2 1010 new page manager creating page

balrajb:


balrajb:
ticket on hold because of new API structure discussed

PR #559, opened 2017-03-13
Wave2 1020 page manager api return menus

balrajb:


balrajb:
remove this old PR as there is a new one here https://github.com/ucalgary/www.ucalgary/pull/560

PR #561, opened 2017-03-15
WAVE2-Fix SSO Varnish Cache Issue

dpfitzsi:
SSO logins have to be done twice because Varnish is caching the login
page. Set a no cache header to prevent this from happening.


dpfitzsi:
There is no JIRA ticket for this yet. Here is the probo to test the header response: https://eb4a6540-b4be-4b88-b5a5-8c74f9195fc4--pr-561.probo.build/

dpfitzsi:
No longer appears to be a need for this. Closing.

PR #565, opened 2017-03-16
Wave2 974 refine degree details CT

nindusaini:


dpfitzsi:
Merged in to wave2-910-912-914-974-fe-compare branch.

PR #587, opened 2017-03-28
WAVE2 1018 Fix $group_name unset / in_array bug

dpfitzsi:


dpfitzsi:
@Robb discussed this today but so we remember in the future, the hook_menu request will check against anonymous access within the callback function and since content ownership is only relevant for authenticated users this was causing the error.

PR #617, opened 2017-04-04
[wave2-1067] compare programs error message

mhughes2012:


robbdavis:
This message is build into the no results section of the view even though it is not supposed to be a no results message. I talked to @jaela but I'm not entirely clear on what the logic behind the 'experiencing tech difficulties' message is supposed to be. 

Hopefully we can just pull it out and replace it with "No Results".. in that case it still needs to be removed from various places. See note in Jirra.

robbdavis:
Mike chatted with Jaela about it and the error message does not appear unless user goes to the page directly via the url without results. It's not ideal to have a "technical difficulties" error in the no results section of the view but that's something we can figure out later if it's an issue.

Approved.

PR #623, opened 2017-04-04
Bug wave2 1032 international admission

TDRoden:


sites/all/themes/ucgy_theme/resources/css/overrides.css
```diff
@@ -192,6 +191,9 @@ select.selectpicker {
   text-transform: none;
   color: #ffffff;
 }
+.bootstrap-select.btn-group .dropdown-menu {
```
gsowrey: Dumb Q: why would we not want this as a globally-available class?

TDRoden: It is a globally available class (?). I'm just changing the target to only restrict dropdown-menus that are within bootstrap-select styled select menus. There are other areas where we might want to use dropdown-menus without a height restriction (like Eric's international brick).

dpfitzsi:
Need to regenerate bricks and overrides css files.

PR #633, opened 2017-04-07
1083 jquery fix to delegate click function even after ajax call

robbdavis:
1083 Make related program drawer work after ajax call on explore programs page.


dpfitzsi:
Applied to http://fgstest-ep2-ucalgary.pantheonsite.io/future-students/graduate/explore-programs and after searching with a keyword the related programs buttons are clickable.

PR #636, opened 2017-04-10
[wave2-1081] Compare Page - when filtering causes no cards to appear …

mhughes2012:
…then an error message should appear — implementation


robbdavis:
I think using views no results message might generally be a better approach for no results messages but I'm not sure how this view is implemented and the method here does work. Peer review approved.

PR #644, opened 2017-04-11
WAVE2-1057 Add preg_match for graduate in URL for breadcrumb

dpfitzsi:


dpfitzsi:
The main difference between the two pieces of markup is most breadcrumbs under /future-students/ will continue to use "Programs Compared". What is changing is that when you go under /future-students/graduate the breadcrumb will change to "Degrees Compared".

PR #666, opened 2017-04-18
WAVE2 1110 Grad details update tuition fees funding section

robbdavis:
WAVE2 1110 Grad details - update tuition fees footer and make update in db.


sites/all/modules/ucalgary/ucgy_fgs/ucgy_fgs.features.field_instance.inc
```diff
@@ -1690,10 +1690,10 @@ function ucgy_fgs_field_default_field_instances() {
     'bundle' => 'fgs_graduate_degree',
     'default_value' => array(
       0 => array(
-        'value' => '<p>Fees listed here are estimates only, and have been rounded to the nearest $100. These amounts should not be considered accurate. All amounts are in Canadian Dollars.<br />
-Fees are subject to change. For more details and the most accurate information go to <a href="http://google.com">Academic Calendar</a><br />
-For more general information, see our <a href="http://google.com">Fees and Funding Guide</a></p>
-',
+        'value' => '<p>Fees listed here are estimates only, and have been rounded to the nearest dollar. These amounts should not be considered accurate. All amounts are in Canadian Dollars.<br />
```
gsowrey: Gonna ask the painful question: Why isn't this a value that comes from the admin page? We've got all that fancy code for storing stuff in variables...

robbdavis: I guess it depends if this information is static across all degrees. If it is, yes, this should be changeable in that fgs admin variables table.

If each program needs the ability to customize this info per degree then it should remain as is.

I'm not sure what the requirement is though... let's ask @Jaela

gsowrey: Confirmed with Jaela. This is the correct implementation. (Ish. I think it could be improved. But time...)

PR #677, opened 2017-04-20
WAVE2 1120 Grad details do not display if no related programs defined

robbdavis:
WAVE2 1120 Related programs section should not display if no related programs displayed in view.


sites/all/modules/ucalgary/ucgy_fgs/templates/node--fgs-graduate-degree.tpl.php
```diff
@@ -194,41 +194,60 @@
...
-        <div class="col-md-12">
-          <h2><?php echo t('Related programs')?></h2>
-          <p><?php echo t('If you\'re interested in this program you might want to explore other programs that UCalgary offers.')?></p>
+<?php // Related Programs Brick
+
+  // The id of this degree
+  $grad_degree_id = $content['field_fgs_grad_degree_grd_prgrm']['#items'][0]['entity']->id;
+
+  // Put the relevent view result in a variable
+  $related_program_view_block2 = views_embed_view('ucgy_fgs_related_programs', 'block_2', $grad_degree_id);
```
gsowrey: Not to be pedantic, but why not have the empty state put out the message below? Why do this in the template?

robbdavis: Because the business wants the entire section (including the header and divider lines) to disappear if there are no results.

robbdavis:
@gsowrey  Thanks for the feedback. I was able to get at both no results conditions in the view template (no strpos!). I also added the view markup to the view template which helps clean up the behemoth graduate_degree.tpl a bit.  Overall, I think this is much cleaner now.

PR #691, opened 2017-04-24
WAVE2 1143 Add message to supervisor list

robbdavis:
WAVE2 1143 Add message to supervisor list


sites/all/modules/ucalgary/ucgy_fgs/templates/views-view--fgs-grad-degree-supervisors.tpl.php
```diff
@@ -61,6 +61,16 @@
     </div>
   <?php endif; ?>
 
+  <?php
+
+    // Check if there are results
+    if (!empty($view->result)) {
+
+      // If there are results, we print out the view footer message (which says there may be more supervisors available)
```
gsowrey: Thought: shouldn't we print out the footer message regardless? If so, should it be external to the view so it's always there?

robbdavis: We have a no results message that printing out, essentially, the same thing. However the markup is slightly different depending on if there is content or not.

We could get rid of the no results message and only print out the footer. Either find markup that works for both or use an statement based on the results. It's essentially the same result as we have now. Do you want me to do it that way or leave it as is?

gsowrey: Let's toss the no results message; it's not serving a purpose if we're putting out the same message in the footer. If we need the message out consistently, regardless of the view results, let's take it out of the view.

robbdavis: I went to pull the messaging out of the view but noticed that the messaging is actually a bit different for results vrs. no results. 

I can hardcode in the messages depending on the results but I'm thinking that just leaving the messaging in the view would allow for on the fly changes on prod if absolutely necessary. (ie... maybe they want to change the url, add a phone number, etc...).

Thoughts?

gsowrey: Hmm. Good question. That could be an accidental oversight. Can you put both versions of the text (and the conditions under which they appear) into the Jira ticket, flag it to Jaela, and let's see if we can whittle down to one?

robbdavis: Should be good now. I've pulled the 'no results' out of the view. The footer will always be displayed no matter what.

PR #702, opened 2017-04-25
Bug wave2 1123 more info spacing

TDRoden:


sites/all/modules/ucalgary/ucgy_gbl/ucgy_gbl_mi/templates/field--field-collection.tpl.php
```diff
@@ -22,11 +22,26 @@
 
 foreach ($items as $item) {
 
-    print '<div class="' . $classes . '"><div class="row media"><div class="col-sm-12">';
+    $spacerClasses;
+    switch (++$count) {
```
gsowrey: I recommend adding $count=0 before the foreach, to ensure that it's proper initialized. Else you could get a situation where $count was previously defined, suddenly polluting this output.

TDRoden: $count=0 has been moved to right before the foreach loop.

PR #710, opened 2017-04-27
Bug wave2 1155 compare tile height

TDRoden:


TDRoden:
Geoff asked me to commit these changes directly to WAVE2-1117, so this PR is no longer needed.

PR #727, opened 2017-05-04
Bug wave2 1175 explore programs filter overlap

TDRoden:
Labels need to be inline-blocks to allow for margins and heights to be
effective.


dpfitzsi:
Code was merged in another PR.

PR #755, opened 2017-05-15
Wave2 1198 google analytics issues

mhughes2012:


sites/all/modules/ucalgary/ucgy_fgs/js/fgs_filters.js
```diff
@@ -200,6 +200,10 @@ function setRelatedEvents() {
             // if the related list already exists, we'll just show/hide it; have to use "hidden" class
             $(parentelem).parent().find('.related-programs-list').toggleClass('hidden');
         }
+
+        // analytics
+        ga_related($(this),parentelem);
```
gsowrey: Introducing a GA dependency into the FGS filters is a risk. Can we not do this any other way?

sites/all/modules/ucalgary/ucgy_fgs/js/fgs_filters.js
```diff
@@ -200,6 +200,10 @@ function setRelatedEvents() {
...
+
         e.stopImmediatePropagation();
```
gsowrey: We had talked about making this "return false" to avoid issues with this cancelling out other functions...?

sites/all/modules/ucalgary/ucgy_fgs/js/fgs_filters.js
```diff
@@ -224,6 +228,7 @@ function clearLabel(filterclass) {
         else {
             $('#filters').find('button').filter('.' + filterclass).removeClass('selected'); // unset the button
             $("#filter-results").find('.label').filter('.' + filterclass).hide(); // hide the filter label
+            ga_clear_filters(filterclass); // analytics
```
gsowrey: Again, a potential risk. Especially if GA doesn't load (as it doesn't if you're logged in), which might cause scripts to break.

sites/all/modules/ucalgary/ucgy_fgs/js/fgs_filters.js
```diff
@@ -266,4 +271,45 @@ function setProgramCompare(parentElem) {
         setLabelEvents();
         updateProgramCount();
     });
-})(jQuery);
\ No newline at end of file
+})(jQuery);
+
+
+/**
```
gsowrey: Let's not embed this in the filters file. This is not an appropriate place for GA code.

sites/all/modules/ucalgary/ucgy_fgs/templates/views-view--fgs-grad-degree-supervisors.tpl.php
```diff
@@ -75,10 +75,10 @@
           $pager = preg_replace('/\<[\/]{0,1}div[^\>]*\>/i', '', $pager);
           print $pager;
         ?>
-        <button class="btn btn-default btn-show-all btn-hide-all" id="hide-button" onClick="window.location.reload()" role="button">Reset</button>
+        <button class="btn btn-default btn-show-all btn-hide-all" id="hide-button" onClick="resetSupervisorList()" role="button">Reset</button>
```
gsowrey: We can't set up an on(click) with jQuery?

sites/all/modules/ucalgary/ucgy_fgs/js/fgs_filters.js
mhughes2012: Calling a GA function here is good. It ensure the tracking fires once at the right time. 
if GA doesn't load (as it doesn't if you're logged in), we capture it and move on. No error in the case where GA is not defined.

sites/all/modules/ucalgary/ucgy_fgs/js/fgs_filters.js
mhughes2012: We did, but we don't need to anymore now that we've moved to functions as opposed to  event listeners here.

sites/all/modules/ucalgary/ucgy_fgs/js/fgs_filters.js
mhughes2012: Same as above.
if GA doesn't load, we capture it and move on. No error in the case where GA is not defined.

sites/all/modules/ucalgary/ucgy_fgs/js/fgs_filters.js
mhughes2012: Sounds good. Moved it to the analytics file.

sites/all/modules/ucalgary/ucgy_fgs/templates/views-view--fgs-grad-degree-supervisors.tpl.php
mhughes2012: Agreed. Was a little lazy here and updated code that was already here. Updated it to an on-click via jQuery now.

PR #759, opened 2017-05-17
[wave1198] typo fix in title of GA event

mhughes2012:


mhughes2012:
Saving this for another day.

PR #763, opened 2017-05-19
1216 Removing extra space from Google Analytics label

robbdavis:
1216 Removing extra space from Google Analytics label


dpfitzsi:
1222 was committed which duplicated this work.

PR #774, opened 2017-05-26
802 related content standardization - subtask of 751

robbdavis:
802 related content standardization - subtask of 721


sites/all/modules/ucalgary/ucgy_gbl/ucgy_gbl_related_content/ucgy_gbl_related_content.module
```diff
@@ -114,4 +127,30 @@ function field_related_link_after_build($form, &$form_state)
...
+ * @param $form_state
+ */
+function make_sure_there_are_two_related_content_field_collections($form, &$form_state){
+
+  // The type of layout selected (two tiles or three tiles)
+  $layout_type = $form_state['entity']->field_ucgy_related_cntnt_layout[LANGUAGE_NONE][0]['value'];
+
+  // The number of tiles that have been added by the user
+  $number_of_fc_items = count($form_state['field']['field_ucgy_related_content'][LANGUAGE_NONE]['entity']);
+
```
robbdavis: In showcase, @geoff mentioned in adding checks for issets to this code. However, the $layout_type will always be set; there is not an option to not have a layout type selected. Likewise, the $number_of_fc_itmes of items is just a count. There will always be a value. So I'm not sure if issets are needed here or if that would just be extra code?

dpfitzsi: Does this mean that the fields are required or that there is a default? I think Geoff's concern is if someone just tried submitting the form we could receive an error.

robbdavis: Even if they just hit submit without filling out anything there is no chance these variables are not set.

taeram:
Closing this PR on GitHub, but keeping the branch around for historical purposes.

PR #775, opened 2017-05-29
1226 New webform template in theme with logic for custom formatting for supervisor node

robbdavis:
1266 New webform template in theme with logic for custom formatting for supervisor node


sites/all/themes/ucgy_theme/templates/webforms/webform-mail.tpl.php
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * The $email['email'] variable can be used to send different e-mails to different users
+ * when using the "default" e-mail template.
+ */
+?>
+
+
+<?php if ($node->type =='gbl_supervisor'){ // we provide special email formatting for the supervisor content type?>
+
+<?php print ($email['html'] ? '<p>' : '') . t('The webform on your supervisor page has been submitted.'). ($email['html'] ? '</p>' : ''); ?>
```
dpfitzsi: Lines 29-59 should be intended since they're within an IF statement. Also there is inconsistent spacing between many of the print statements, limit it to one line rather than sometimes using two lines

sites/all/themes/ucgy_theme/templates/webforms/webform-mail.tpl.php
```diff
@@ -0,0 +1,80 @@
...
+<?php print ($email['html'] ? '<p>' : '') . t('[submission:values:why_do_you_want_to_work_with_me]') . ($email['html'] ? '</p>' : ''); ?>
+
+
+<?php print ($email['html'] ? '<p><b>' : '') . t('Additional message:') . ($email['html'] ? '</b></p>' : ''); ?>
+
+<?php print ($email['html'] ? '<p>' : '') . t('[submission:values:additional_message]') . ($email['html'] ? '</p>' : ''); ?>
+
+
+
+<?php } else { // otherwise it's just the default email formatting?>
```
dpfitzsi: Lines 61 to 78 should be intended within the ELSE statement

sites/all/themes/ucgy_theme/templates/webforms/webform-mail.tpl.php
robbdavis: Hey Dan, I know it looks odd but this was the only way to get the plain text email to format correctly. The returns and indents actually show up in the email.

sites/all/themes/ucgy_theme/templates/webforms/webform-mail.tpl.php
robbdavis: As above, the indented lines actually render in the email as indented. This format is copied exactly from the webform-mail.tpl.php in the webform module.

I think that eventually implementing HTML emails is the way around this and would be a good improvement for the site in general.

dpfitzsi: Ah I had this feeling in the back of my mind. I wasn't sure if the indenting problem only effected HTML emails. Ok we will leave things as is.

PR #784, opened 2017-06-08
Front end font size change

TDRoden:
This is a test build for heading font-size change evaluation. Based on Mariana and other's feedback this will become an actual change request.


TDRoden:
Replaced with an actual improvement branch: WSO-142_header-font-size-increase

PR #790, opened 2017-06-12
WSO 63 google analytics tabbing

TDRoden:


sites/all/themes/ucgy_theme/resources/js/ucgy_analytics.js
```diff
@@ -36,6 +36,12 @@ var ucgy_analytics;
             // We had to duplicate this code in order to manage secondary trackers more effectively
             // Catch the closest surrounding link of a clicked element.
             $(document.body).bind("mousedown keyup touchstart", function(event) {
+
+                // If it is a key press, we are only concerned with tracking 'enter' key interactions
+                if ((event.type == "keyup") && (event.which != 13)) {
+                    return;
+                }
+
```
SandyVallee: where is the code that does throw the ga event if its an enter to launch the link?  should there be an else statement to this?

SandyVallee: I see that it is still launching the ga event on an enter key press so I'm sure its somewhere else in the file=) You can ignor this.

TDRoden: If it is an 'enter' keyboard event, it just continues through to the next block and is processed along with the mousedown and touchstart events. All other keyboard events are discarded/ignored. I feel an else statement here complicates things as you'd also have to account for mousedown and touchstart.

SandyVallee: sounds good =)

PR #796, opened 2017-06-16
Wso 3 ga fix international js

SandyVallee:
testing probo


sites/all/modules/ucalgary/ucgy_gbl/ucgy_gbl_countries/js/ucgy_gbl_international.js
```diff
@@ -74,8 +77,10 @@
         });
 
         // hide country dropdown
+        var itemClosing = '';
         $(this).on('hide.bs.dropdown', function(e) {
             country_name = e.relatedTarget.text;
+            itemClosing = '#' + e.relatedTarget.id;
```
taeram: The `var itemClosing = ''` above should be removed as it doesn't actually do anything because the variable is only created and used inside this function.

To set the `itemClosing` variable scope so it's only active inside the function, instead prefix this line with `var`:
```js
var itemClosing = '#' + e.relatedTarget.id;
```

taeram:
Closing PR as WSO-3 has been cancelled

PR #820, opened 2017-07-25
WSO-272 - Remove CRUMBS

SandyVallee:
Removed crumbs module and directory. Removed references to it in any UCGY files.  Did not touch css or scss as Terry mentioned that it is still called crumbs but is used to house other content now.  I also did not touch too much in the contributed section, as I thought that it would be overwritten with contrib module updates anyways. I also did not do a code clean due to the type of files touched and most things were whole file deletions.  If I need to do one on any of these files please let me know what needs to be cleaned.

Thanks Sandy


taeram:
@SandyVallee I agree with @TDRoden that we should revert the change to `sites/all/modules/contrib/apachesolr/CHANGELOG.txt`. 

Looks good other than that.

PR #819, opened 2017-07-24
Wso 266 lock down access

christophergraha1:
Permission updates to people config pages

PR #827, opened 2017-08-03
WSO-286 - Code cleanup for WSO-266

taeram:
This branch cleans up code as requested by @gsowrey in [WSO-266](https://github.com/ucalgary/www.ucalgary/pull/819).

This code cleanup contains some rather major changes.

Addition of code to be used across all UCalgary modules:
* Addition of a `ucgy_core_constants.inc` file, containing a list of constant variables 
   * Current constants include roles, module names, and package names
* Addition of a `ucgy_core_helpers.inc` file, containing a set of helper functions
   * Generic functions that are used across multiple modules should be moved here

Refactoring of existing functions, and addition of new helper functions:
* Renamed `_is_first_fc_item()` function to `ucgy_fc_item_is_first()` 
* Replaced ` has_admin_privileges()` function with `ucgy_user_is_superadmin()` to be more explicit
* Added `ucgy_environment_is_pantheon()` to test if we're on Pantheon
* Added `ucgy_user_has_roles($user, $roles)` function

Removed unused / deprecated functions:
* `_is_first_fc_item()`
* `is_brick_type()`
* `_get_array_value_by_pattern()`
* `_ucgy_core_get_faculties()`

Other improvements:
* Tidied a lot of comments, making them more specific, or removing them entirely if they were trivial / not useful.
* Sorted `case` statements in a `switch` to be alphabetical, and therefore easier to find while scanning the code
* Removed commented out code


sites/all/modules/ucalgary/ucgy_core/ucgy_core_roles/ucgy_core_roles.module
```diff
@@ -14,79 +14,52 @@ include_once 'ucgy_core_roles.features.inc';
...
   static $drupal_static_fast;
-  if (!isset($drupal_static_fast['perm'])) {
+  if (!isset($drupal_static_fast)) {
     $drupal_static_fast['perm'] = &drupal_static('user_access');
   }
-  $perm = &$drupal_static_fast['perm'];
 
-  if (isset($perm[$user->uid])) {
+  // If we're on Pantheon live environment, remove the Developer role from this user
+  if (isset($drupal_static_fast['perm'][$user->uid]) && is_pantheon_live_environment()) {
```
taeram: Should we even have this `is pantheon` check here? Is the Developer role necessary on non-Live servers?

sites/all/modules/ucalgary/ucgy_core/ucgy_core_roles/ucgy_core_roles.module
```diff
@@ -138,12 +106,6 @@ function ucgy_core_roles_form_alter(&$form, &$form_state, $form_id) {
   global $user;
 
   switch ($form_id) {
-    case "page_manager_page_form_menu":
```
taeram: Moved these case statement down lower so the entire `switch` statement is alphabetized.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_roles/ucgy_core_roles.module
```diff
@@ -157,59 +119,37 @@ function ucgy_core_roles_form_alter(&$form, &$form_state, $form_id) {
...
       }
 
-      //this was removing ability to reuse the ffp entity
-      // hide($form['additional_settings']);
       // We are going to set the default value of category
       $form['reusable']['category']["#default_value"] = 'Existing bricks';
       // then hide the category field from the user
       hide($form['reusable']['category']);
       break;
-    case "views_form_bulk_update_groups_and_roles_page_1";
```
taeram: Removed as per Geoff's request.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_roles/ucgy_core_roles.module
```diff
@@ -265,45 +191,22 @@ function ucgy_core_roles_form_alter(&$form, &$form_state, $form_id) {
 function ucgy_core_roles_menu_local_tasks_alter(&$data, $router_item, $root_path) {
   global $user;
 
-  // If the user is User 1, we're allowing everything
-  if ($user->uid === "1") {
+  if (is_superadmin($user)) {
     return TRUE;
   }
 
-  /* THIS IS A TEMPORARLY BLOCK ON THE DEVELOPMENT GROUP; TO BE REMOVED */
```
taeram: Removed as per Geoff's request.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_roles/ucgy_core_roles.module
```diff
@@ -312,251 +215,135 @@ function ucgy_core_roles_menu_local_tasks_alter(&$data, $router_item, $root_path
...
- * Implements hook_menu_contextual_links_alter().
- *
- * This is being used to turn off contextual links in places where users
- * shouldn't see them.
- *
- * @param $links
- * @param $router_item
- * @param $root_path
- */
-function ucgy_core_roles_menu_contextual_links_alter(&$links, $router_item, $root_path) {
```
taeram: Removed as per Geoff's request.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_roles/ucgy_core_roles.module
```diff
@@ -312,251 +215,135 @@ function ucgy_core_roles_menu_local_tasks_alter(&$data, $router_item, $root_path
...
-
-/**
- * Implements hook_admin_menu_output_alter().
- *
- * Being used to hide the admin menu items (except flush caches) for certain
- * roles. Temporary.
- *
- * @param $content
- */
-function ucgy_core_roles_admin_menu_output_alter(&$content) {
```
taeram: Removed as per Geoff's request.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_roles/ucgy_core_roles.module
```diff
@@ -312,251 +215,135 @@ function ucgy_core_roles_menu_local_tasks_alter(&$data, $router_item, $root_path
...
 /**
- * @param StdClass $user Should be the Drupal $user object
+ * Does the user have access to the admin/people page?
  *
- * @return bool True if the user has admin powers, or false
+ * @param array $arg The argument
+ *
+ * @return bool
  */
-function has_admin_privileges($account = NULL) {
```
taeram: Refactored into the `is_superadmin()` function.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_roles/ucgy_core_roles.module
```diff
@@ -312,251 +215,135 @@ function ucgy_core_roles_menu_local_tasks_alter(&$data, $router_item, $root_path
...
-    // If there's an intersection of the two arrays, user has admin privs
-    $matches = array_intersect(array_values($user->roles), $admin_roles);
-    return !empty($matches);
-  }
-  return FALSE;
-
-}
-
-// Custom access callback used in hook_menu_item_alter above
-function ucgy_core_roles_access_callback($arg) {
```
taeram: Removed as it's unused.

sites/all/modules/ucalgary/ucgy_core/ucgy_core.module
```diff
@@ -167,80 +140,6 @@ function ucgy_core_tokens($type, $tokens, array $data = array(), array $options
...
- * @param FieldablePanelsPaneEntity $fpp
- *   The object we're testing
- * @return bool
- *   Is the object a brick or not?
- *
- * TODO: Need to make this more generic, so we can have N "hero" panels that use the same classification
- * TODO: We need a static of brick types that match the machine names, so we have consistency
- * TODO: Not all bricks are FPPs
- */
-function is_brick_type($type = null, $fpp = null)
```
taeram: Removed as it's unused.

sites/all/modules/ucalgary/ucgy_core/ucgy_core.module
```diff
@@ -167,80 +140,6 @@ function ucgy_core_tokens($type, $tokens, array $data = array(), array $options
...
- * Function to find key patterns within arrays.
- *
- * @param string $pattern
- *   The pattern we're looking for inside the array.
- * @param array $input
- *   The array to search.
- * @return array
- *   The results of the search.
- */
-function _get_array_value_by_pattern($pattern, $input)
```
taeram: Removed as it's unused.

sites/all/modules/ucalgary/ucgy_core/ucgy_core.module
```diff
@@ -167,80 +140,6 @@ function ucgy_core_tokens($type, $tokens, array $data = array(), array $options
...
- * @param string $needle
- *   What we're looking for...
- * @param array $haystack
- *   Where it's hiding...
- * @param bool $strict
- *   We might need to make the search strict (like with in_array() ).
- * @return bool
- *   Did we find it or not?
- */
-function in_array_r($needle, $haystack, $strict = false)
```
taeram: Moved to `ucgy_core_helpers.inc`

sites/all/modules/ucalgary/ucgy_core/ucgy_core.module
```diff
@@ -658,17 +557,3 @@ function _ucgy_core_get_fpp_from_subtype($subtype = null) {
 
   return $object;
 }
-
-function _ucgy_core_get_faculties() {
```
taeram: Removed as it's unused.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_roles/ucgy_core_roles.module
```diff
@@ -265,45 +191,22 @@ function ucgy_core_roles_form_alter(&$form, &$form_state, $form_id) {
...
 
-  // If the user is User 1, we're allowing everything
-  if ($user->uid === "1") {
-    return TRUE;
+  if (user_is_superadmin($user)) {
+    return;
   }
 
-  /* THIS IS A TEMPORARLY BLOCK ON THE DEVELOPMENT GROUP; TO BE REMOVED */
-  if (in_array('Development', $user->roles)) {
```
taeram: Removed as requested by Geoff

sites/all/modules/ucalgary/ucgy_core/ucgy_core_roles/ucgy_core_roles.module
```diff
@@ -312,251 +215,135 @@ function ucgy_core_roles_menu_local_tasks_alter(&$data, $router_item, $root_path
...
 }
 
 /**
  * Implements hook_menu().
+ *
  * Builds the menu for use in admin toolbar
  */
-
 function ucgy_core_roles_menu() {
+  return [
```
taeram: Reformatted this using the short array syntax (`array('foo') => ['foo']`)

sites/all/modules/ucalgary/ucgy_core/ucgy_core_constants.inc
```diff
@@ -0,0 +1,41 @@
...
+const UCGY_PACKAGE_UNIVERSITY_OF_CALGARY_GLOBAL = 'University of Calgary (Global)';
+
+/**
+ * Site Roles
+ */
+const UCGY_ROLE_ADMINISTRATOR = 'administrator';
+const UCGY_ROLE_CONTENT_ADMINISTRATOR = 'Content Administrator';
+const UCGY_ROLE_CONTENT_EDITOR = 'Content Editor';
+const UCGY_ROLE_DEVELOPER = 'Developer';
+const UCGY_ROLE_DEVELOPMENT = 'Development';
```
gsowrey: Remove; this is not a role we want in our system.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_constants.inc
```diff
@@ -0,0 +1,41 @@
...
+const UCGY_ROLE_FUTURE_GRAD_STUDENTS_EDIT_SUPERVISOR = 'FGS - Edit Supervisor';
```
gsowrey: This should be moved to the ucgy_fgs module, in a new ticket. This role's existence should also be bumped to a spike for removal -- it's a big issue/risk for us.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_constants.inc
```diff
@@ -0,0 +1,41 @@
...
+const UCGY_ROLE_FUTURE_GRAD_STUDENTS_SUPERVISOR = 'FGS - Supervisor';
```
gsowrey: Move to ucgy_fgs, in a new ticket

sites/all/modules/ucalgary/ucgy_core/ucgy_core_constants.inc
```diff
@@ -0,0 +1,41 @@
...
+const UCGY_ROLE_GLOBAL_SITE_ALERT = 'Global Site Alert';
```
gsowrey: We need this role, but only on the www site; not sure what's the best module for that. We might need a new one to deal with www-only modules. Suggest new ticket to move this one.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_constants.inc
```diff
@@ -0,0 +1,41 @@
...
+const UCGY_ROLE_GROUP_ADMINISTRATOR = 'Group Administrator';
```
gsowrey: This role should only exist in content ownership; leave here for now, but we should comment/flag it for moving later on.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_constants.inc
```diff
@@ -0,0 +1,41 @@
...
+const UCGY_ROLE_MARKETING = 'Marketing';
```
gsowrey: We should remove this one. Honestly not sure why it's here.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_constants.inc
```diff
@@ -0,0 +1,41 @@
...
+const UCGY_ROLE_NEWS_AND_EVENTS_FEEDS_IMPORTER = 'News and Events Feeds Importer';
```
gsowrey: Move to www-only module in a separate ticket

sites/all/modules/ucalgary/ucgy_core/ucgy_core_constants.inc
```diff
@@ -0,0 +1,41 @@
...
+const UCGY_ROLE_SITE_ADMINISTRATOR = 'Site Administrator';
+const UCGY_ROLE_STRATEGIC_COMMUNICATIONS = 'Strategic Communications';
```
gsowrey: Not sure if we should remove this one, or move to www-only module. Where is this used?

sites/all/modules/ucalgary/ucgy_core/ucgy_core_constants.inc
```diff
@@ -0,0 +1,41 @@
...
+const UCGY_ROLE_SUPPORT = 'Support';
+
+/**
+ * Module names
+ */
+const MODULE_UCGY_CORE_ROLES = 'ucgy_core_roles';
+const MODULE_FIELDABLE_PANELS_PANE = 'fieldable_panels_pane';
```
gsowrey: I get the previous one, but why this and the next? Shouldn't these be required modules? Not convinced we need constants for this.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_constants.inc
```diff
@@ -0,0 +1,41 @@
...
+const MODULE_SIMPLESAMLPHP_AUTH = 'simplesamlphp_auth';
+
+/**
+ * Pantheon Environments
+ */
+const ENVIRONMENT_PANTHEON_LIVE = 'live';
```
gsowrey: Only the live env? Not dev, test, staging, or the local Vagrant?

sites/all/modules/ucalgary/ucgy_core/ucgy_core_helpers.inc
```diff
@@ -0,0 +1,79 @@
...
+}
+
+/**
+ * Are we on the Pantheon environment?
+ *
+ * @param string $environmentName The pantheon environment name. Optional
+ *
+ * @return bool
+ */
+function environment_is_pantheon($environmentName = null) {
```
gsowrey: Since we're checking the $_ENV variable anyway for a very specific string that wouldn't exist elsewhere, moving the comparison to a constant is of limited value, IMO. 

Also, as we're trying to get _off_ Pantheon, not sure how long a method like this will need to exist.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_helpers.inc
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Is this user a superadmin (e.g. user 1) or have a Support role?
+ *
+ * @param object $user The user object
+ *
+ * @return bool
+ */
+function user_is_superadmin($user) {
+  return ((isset($user->uid) && $user->uid === "1") || user_has_roles($user, UCGY_ROLE_SUPPORT));
```
gsowrey: When using $user within Drupal methods, it's a good practice to make sure you've got the full user data, just in case: 

if (isset($user->uid)) {
  $account = user_load($user->uid);
  if ((int)$account->uid === int(1) || user_has_roles($account, UCGY_ROLE_SUPPORT))...

sites/all/modules/ucalgary/ucgy_core/ucgy_core_helpers.inc
```diff
@@ -0,0 +1,79 @@
...
+ * @param array $roles The user roles
+ *
+ * @return bool
+ */
+function user_has_roles($user, $roles) {
+  if (!isset($user->roles)) {
+    return false;
+  }
+
+  if (is_array($roles)) {
```
gsowrey: So what happens if $roles isn't an array? According to line 54 (above), you can also expect just a string. Might want to do an array check first, and reset for an array if needed.

sites/all/modules/ucalgary/ucgy_gbl/ucgy_gbl_tabs/templates/field-collection-item--field-tab-item.tpl.php
```diff
@@ -29,7 +29,7 @@
  * @see template_process()
  */
 //kpr($content);
-$active = (_is_first_fc_item($field_collection_item, $field_collection_item->item_id, 'field_uctab_item')) ? " active" : "";
```
gsowrey: _is_first_fc_item was originally meant to be a globally-useful method, regardless of which module used it. So moving it to ucgy_gbl_tabs might be a problem in the future, if we either a) need to move it back, or b) someone makes another helper function to do the same thing.

sites/all/modules/ucalgary/ucgy_gbl/ucgy_gbl_tabs/ucgy_gbl_tabs.module
```diff
@@ -103,4 +103,33 @@ function ucgy_gbl_tabs_after_build($form, &$form_state) {
   $form['field_uctab_item'][LANGUAGE_NONE][2]['field_uctab_cta'][LANGUAGE_NONE][0]['title']['#description'] = t('When possible try to use action-oriented type language.<br/>  Should be no more than two to five words.');
 
   return $form;
-}
\ No newline at end of file
+}
+
+/**
+ * function ucgy_gbl_tabs_is_first_fc_item().
```
gsowrey: Er, yeah, what I said above...

sites/all/modules/ucalgary/ucgy_core/ucgy_core_helpers.inc
gsowrey: Oh, and one more thing. Do we have a standard naming convention for helper functions? My fear is that a lack of a suitable pattern might a) cause collisions with third-party modules, or b) we'll see everything under the sun for patterns (ew).

sites/all/modules/ucalgary/ucgy_core/ucgy_core_constants.inc
taeram: It looks like we don't reference it anywhere. Removed!

sites/all/modules/ucalgary/ucgy_core/ucgy_core_constants.inc
taeram: I had envisioned this list as a set of constants we could use in our code when we need the names of our ucgy_* modules, or the names of contrib modules, in order to prevent spelling mistakes in our code when just using plain strings.

With that, this list would eventually grow to encompass any module we reference in our ucgy_* code

sites/all/modules/ucalgary/ucgy_core/ucgy_core_constants.inc
taeram: So far, we don't test for anything except for the live environment, so I didn't think it was necessary to add extra environments.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_helpers.inc
taeram: Currently, we have 11 places in our code where we test if the environment is Pantheon, so I think this function does have some value in bringing consistency and improving readability.

However, we only test if we're on the "live" environment in two places: in the roles code in this pull request, and in `ucalgary_redirects.inc` where we figure out proxies. I don't know if we want / need a specific test for Pantheon live in the roles module, as it seems odd to have non-setup code that only runs on Live Pantheon. If we were to remove the Pantheon Live test from the roles module, we could safely remove the `$environmentName` input parameter from this function, and just have it test whether we're on Pantheon or not.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_helpers.inc
taeram: Whoops! This was an error on line 54, and has been corrected. The `user_has_roles()` function does expect that `$roles` is an array. I've also added a type hint to `user_has_roles()` to this effect.

sites/all/modules/ucalgary/ucgy_gbl/ucgy_gbl_tabs/templates/field-collection-item--field-tab-item.tpl.php
taeram: I've moved it back to the `ucgy_core_helpers.inc` file, and removed the `ucgy_gbl_tabs_` prefix from the function.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_helpers.inc
taeram: I really struggled with that question when writing these helpers, which is why I left them without prefixes.

The most Drupal-esque way to add the prefixes would be to add the module name as a prefix, so `environment_is_pantheon()` would become `ucgy_core_environment_is_pantheon()`. However, I'd like to have them namespaced, but still global, so I've trimmed that down a bit, and updated the functions to just have a `ucgy_` prefix, making this `ucgy_environment_is_pantheon()`.

taeram: As to the names of the functions themselves, I like `ucgy_<thing being tested>_<verb>_<thing we're looking for>`. So checking for a user role becomes `ucgy_user_has_roles()`, checking the pantheon environment becomes `ucgy_environment_is_pantheon()`.

To this effect, the `ucgy_is_first_fc_item()` function could be renamed `ucgy_fc_item_is_first()`.

sites/all/modules/ucalgary/ucgy_core/ucgy_core.module
```diff
@@ -3,37 +3,10 @@
...
- * This is a simple test to determine if a given FC item is the first in a Field Collection.
- *
- * @param FieldCollectionItemEntity $field_collection_item
- *   This is the field collection item being checked; it's used to get its parent.
- * @param integer $fc_item_id
- *   The ID of the first item; we need to check the FC item against this.
- * @return bool
- *   It either is or isn't the first item.
- */
-function _is_first_fc_item($field_collection_item = null, $fc_item_id = 0, $fc_obj = null)
```
taeram: Moved to helpers file.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_constants.inc
gsowrey: Hmm. I'm not convinced this will work as well as you might think, and if only because I've never seen it as a pattern in the Drupal ecosystem (ie. it's not a current pattern or expectation). Maybe let's skip the module constant names for now, unless we can start seeing real use for it?

taeram: Sounds good. I've removed the non-ucgy module constants from the code.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_helpers.inc
```diff
@@ -14,9 +14,9 @@
  *
  * @return bool
  */
-function in_array_r($needle, $haystack, $strict = FALSE) {
+function ucgy_in_array_r($needle, $haystack, $strict = FALSE) {
   foreach ($haystack as $item) {
-    if (($strict ? $item === $needle : $item == $needle) || (is_array($item) && in_array_r($needle, $item, $strict))) {
+    if (($strict ? $item === $needle : $item == $needle) || (is_array($item) && ucgy_in_array_r($needle, $item, $strict))) {
```
gsowrey: Okay, this is going to be annoying as hell, but I'll offer a thought: I think that in_array_r could remain as-is, as it's (kind of) an extension of the PHP in_array method. (I stole the method definition from Stackoverflow. I take zero credit for the idea.) It's similar to the existing one, but offers a bit more. Just my $0.05.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_helpers.inc
```diff
@@ -27,17 +27,31 @@ function in_array_r($needle, $haystack, $strict = FALSE) {
 /**
  * Are we on the Pantheon environment?
  *
- * @param string $environmentName The pantheon environment name. Optional
+ * @return bool
+ */
+function ucgy_environment_is_pantheon() {
+  return isset($_ENV["PANTHEON_ENVIRONMENT"]);
```
gsowrey: Nice! :)

sites/all/modules/ucalgary/ucgy_core/ucgy_core_helpers.inc
```diff
@@ -27,17 +27,31 @@ function in_array_r($needle, $haystack, $strict = FALSE) {
...
+ * @param integer $fc_item_id The ID of the first item; we need to check the FC item against this.
+ * @param string $fc_obj The field collection object name
  *
  * @return bool
  */
-function environment_is_pantheon($environmentName = null) {
-  if (isset($_ENV["PANTHEON_ENVIRONMENT"])) {
-    if ($environmentName !== NULL) {
-      return $_ENV["PANTHEON_ENVIRONMENT"] === $environmentName;
+function ucgy_fc_item_is_first(\FieldCollectionItemEntity $field_collection_item, $fc_item_id, $fc_obj) {
```
gsowrey: Gonna claim stupid here, but what does the "\" in front of FieldCollectionItemEntity do? Never seen it before (in PHP 5.x, anyway...)

sites/all/modules/ucalgary/ucgy_core/ucgy_core_roles/ucgy_core_roles.module
```diff
@@ -318,9 +318,9 @@ function admin_people_callback($arg) {
  * Add Publish/Unpublish link for Fieldable Panels Panes for superadmins or Administrators
  */
 function ucgy_core_roles_contextual_links_view_alter(&$element, $items) {
-  if (isset($element['#element']['#entity_type']) && $element['#element']['#entity_type'] === MODULE_FIELDABLE_PANELS_PANE) {
+  if (isset($element['#element']['#entity_type']) && $element['#element']['#entity_type'] === 'fieldable_panels_pane') {
```
gsowrey: Oooooh, actually, now I'm really glad that I suggested removing that constant. 'Cuz the module name doesn't always equal the type of data you're looking at (thanks, Open Source Community...)

sites/all/modules/ucalgary/ucgy_core/ucgy_core_helpers.inc
taeram: When referencing a class, it's important to include the correct namespace for that class. `\FieldCollectionItemEntity` basically says "we expect the FieldCollectionItemEntity class to be found in the root namespace".

This will become useful practice when we start using namespaced classes. If this function were contained in a namespaced class (e.g. `\Ucgy\Helpers::fc_item_is_first()`), specifying the relative namespace (e.g. leaving out the leading slash) would cause PHP to look for the `FieldCollectionItemEntity` class in the current namespace (`\UCGY\Helpers\FieldCollectionItemEntity`) which is incorrect. By adding the leading slash in front of `FieldCollectionItemEntity`, we tell PHP to look only in the root namespace for it.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_helpers.inc
taeram: I agree that simple is best, but I'd like to keep the `ucgy_` prefix on this so there's no confusion as to where this function lives. If we left it as `in_array_r`, I think it'd be easy to assume it was a built-in PHP function, and waste a few minutes here and there searching the php.net documentation for it.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_roles/ucgy_core_roles.module
taeram: In that case, should we remove the `===` and instead just use `==` when testing the module name?

gsowrey:
Huh. Well, I’ll be. Didn’t know that worked in PHP 5.x. Sweet.

gsowrey:
Fair point.

gsowrey:
Counter-question: Should we be sticking to “===” whenever possible? In this case, it’s certainly an option (as it is a string compare)

sites/all/modules/ucalgary/ucgy_core/ucgy_core_roles/ucgy_core_roles.module
taeram: Re your question "Counter-question: Should we be sticking to “===” whenever possible? In this case, it’s certainly an option (as it is a string compare)", yes, we already have the type safe comparison as a recommendation in our coding standards: https://github.com/ucalgary/drupal-coding-standards-and-processes/blob/master/PHP.StyleGuide.md#12-operators
 - See the third sentence

So changing this to a `==` would go against our coding standards. Maybe we want to either a) cast everything to a string and use `===`, or b) create a helper function to do this string-only comparison for us.

taeram: I've cast'ed the `entity_type` to a string so we can use `===` in this case without worry.

taeram:
@gsowrey Sending back to you for a final review.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_helpers.inc
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * A multidimensional version of in_array.
+ *
+ * @param string $needle What we're looking for...
+ * @param array $haystack Where it's hiding...
+ * @param bool $strict We might need to make the search strict (like with in_array() ).
+ *
+ * @return bool
+ */
+function ucgy_in_array_r($needle, $haystack, $strict = FALSE) {
```
gsowrey: Thought: Should we start generating PHPDocs, so that these helpers are known, rather than tucked away somewhere?

sites/all/modules/ucalgary/ucgy_core/ucgy_core_roles/ucgy_core_roles.features.user_role.inc
```diff
@@ -8,49 +8,34 @@
...
+    ],
+    'Content Editor' => [
+      'name' => 'Content Editor',
+      'weight' => 4,
+    ],
+    'Developer' => [
+      'name' => 'Developer',
+      'weight' => 2,
+    ],
+    'FGS - Supervisor' => [
```
gsowrey: This needs to be in ucgy_fgs, not here.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_roles/ucgy_core_roles.features.user_role.inc
```diff
@@ -8,49 +8,34 @@
...
+      'name' => 'FGS - Supervisor',
+      'weight' => 13,
+    ],
+    'Group Administrator' => [
```
gsowrey: Should probably @TODO this one, as we'll be moving it to the new Content Ownership module when it's done.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_roles/views/ucgy_core_roles.views.inc
```diff
@@ -7,17 +7,13 @@
  * @return array hook data
  */
 function ucgy_core_roles_views_data() {
-    $data = array();
-
-    // Nothing yet! So we have a placeholder...
-
-    return $data;
+  return [];
```
gsowrey: Gonna (sorta) call you on this one. Why the [] short form instead of array()? Does this follow standards (not ours; the industry)?

sites/all/modules/ucalgary/ucgy_core/ucgy_core_helpers.inc
taeram: I've never actually had an instance where PHPDocs were useful to members of a team. They're often useful to external people (e.g. if we had someone write code for us), but on a team, we would a) encourage our fellow developers to read through the helpers file, and keep it organized so it's easy to review, and b) if there's something that could / should be done using a helper function, we will typically catch it in a code review and request the change then.

If we end up with a massive library of helper functions, maybe a quick reference document would be useful (e.g. the list of functions and a one liner of what they do), but for the time being I think external documentation would be overkill.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_roles/views/ucgy_core_roles.views.inc
taeram: Yep, this is in our UCalgary coding standards, as well as the Drupal standards: https://www.drupal.org/docs/develop/standards/coding-standards#array
> Arrays should be formatted using short array syntax with a space separating each element (after the comma), and spaces around the => key association operator, if applicable

sites/all/modules/ucalgary/ucgy_core/ucgy_core_roles/ucgy_core_roles.features.user_role.inc
taeram: I've TODO'd it.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_roles/ucgy_core_roles.features.user_role.inc
taeram: Done!

* Moved `FGS - Supervisor` to the `ucgy_fgs` module, and adding the `Group Administrator` TODO comment: 4aa9949404103c62322a90a93b5d91690cea1b43
* Removing the accidentally committed `var_dump()`: f7fad410ca0b6ec50b9dfea780be9fbc9eac8f24

sites/all/modules/ucalgary/ucgy_gbl/ucgy_gbl_tabs/templates/field-collection-item--field-tab-item.tpl.php
```diff
@@ -29,7 +29,7 @@
  * @see template_process()
  */
 //kpr($content);
```
dpfitzsi: This is super nitpicky but we should probably remove debug messages completely.

PR #834, opened 2017-08-18
WSO-233 - Update Variable Table

SandyVallee:
So we have to set a default email as this module uses tokens and tokens requires a default to be set.  It never gets sent to though due to the validation in the form, but the code is necessary so the entered in email into the variable table will send.


sites/all/modules/ucalgary/ucgy_gbl/ucgy_gbl_supervisor/ucgy_gbl_supervisor.module
```diff
@@ -256,7 +256,7 @@ function ucgy_gbl_supervisor_tokens($type, $tokens, array $data = array(), array
       switch ($name) {
         // set the token value
         case 'gbl-supervisor-change-email':
-          $replacements[$original] =  variable_get('fgs_email', 'jaela.curtis@ucalgary.ca');
+          $replacements[$original] =  variable_get('fgs_email', 'noreply@ucalgary.ca');
```
taeram: We should make this a constant, as it's a plain string that we use in several parts of our code.

Please put the noreply email address in a new file at `sites/all/modules/ucalgary/ucgy_gbl/ucgy_gbl_constants.inc` as a constant:
```php
<?php
const UCGY_EMAIL_NOREPLY = 'noreply@ucalgary.ca';
```

Then, change all the code in this pull request to use the new constant:
```php
variable_get('fgs_email', UCGY_EMAIL_NOREPLY);
```

PR #836, opened 2017-08-22
WSO-225 - Implement Auto Complete on the Graduate Explore Programs Page

dpfitzsi:
Fixed duplication issue.

ucgy_core_autocomplete module needs to be enabled to test this.


sites/all/modules/ucalgary/ucgy_core/ucgy_core_autocomplete/ucgy_core_autocomplete.module
```diff
@@ -0,0 +1,123 @@
...
+      break;
+    case 'solr':
+      $view_name = 'fgs_solr_results';
+      $display_id = 'block';
+      $results = views_get_view_result($view_name, $display_id);
+      foreach ($results as $result) {
+        $matches[$result->degree_name] = $result->degree_name;
+      }
+      break;
+    case 'default':
```
christophergraha1: Not sure using case here makes a lot of sense over a standard condition??

sites/all/modules/ucalgary/ucgy_core/ucgy_core_autocomplete/ucgy_core_autocomplete.module
```diff
@@ -0,0 +1,123 @@
...
+        $keywords = explode(' ', $item->title);
+        $matches[$keywords[0]] = $keywords[0];
+      }
+
+      foreach ($result as $item) {
+        $keywords = explode(' ', $item->keyword_field);
+        $matches[$item->keyword_field] = $keywords[0];
+      }
+
+      $result = db_query_range('SELECT n.field_gbl_supervisor_last_name_value AS keyword_field FROM {field_data_field_gbl_supervisor_last_name} n WHERE n.field_gbl_supervisor_last_name_value LIKE :string', 0, 10, [
```
taeram: From this query onwards, it looks like we're requesting the same sort of data, and parsing it the same way as well.

If we could so something like put the SQL in an array, and iterate through that array instead of having four individual sections, it'd improve the robustness of the code (fewer places to make small errors), and also improve readability.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_autocomplete/ucgy_core_autocomplete.module
```diff
@@ -0,0 +1,123 @@
...
+
+      foreach ($result as $item) {
+        $keywords = explode(' ', $item->keyword_field);
+        $matches[$item->keyword_field] = $keywords[0];
+      }
+
+      break;
+  }
+
+  drupal_json_output(array_unique($matches));
```
taeram: From context, I assume this prints the JSON headers and JSON string to the page, but it'd be nice to have a one line comment above it just explaining that to improve readability.

sites/all/modules/ucalgary/ucgy_fgs/js/ucgy_fgs.js
```diff
@@ -88,10 +61,10 @@
 
 
 
-      /**
-     * Update delivery fields available bases on delivery type value
-     * @param dtype delivery type
-     */
+    /**
+    * Update delivery fields available bases on delivery type value
+    * @param dtype delivery type
```
taeram: Should have a hard return between the function description and `@param` declaration. 

It also looks like `dtype` is a string, so that would make the `@param` into: 
```
* @param string dtype The delivery type
```

dpfitzsi:
Now that there is 489 we don't need this branch.

PR #839, opened 2017-09-05
WSO-168 Add access check for menu items (403 Forbidden)

dpfitzsi:
Need to create a new page manger page for 403 forbidden with no content. Add the url that was just generated to 403 access denied field on the Site Information screen and everything will be set.


sites/all/themes/ucgy_theme/includes/alter.inc
```diff
@@ -52,7 +52,7 @@ function ucgy_theme_js_alter(&$javascript)
 
     // Load proper jQuery version on 403 and 404 pages
     $status = drupal_get_http_header('Status');
-    if ($status = '403 Forbidden' || $status = '404 Not Found') {
+    if ($status === '403 Forbidden' || $status === '404 Not Found') {
```
dpfitzsi: Good catch! Thanks.

sites/all/themes/ucgy_theme/includes/preprocess.inc
```diff
@@ -33,21 +33,26 @@ function ucgy_theme_preprocess_html(&$variables)
...
     // Adding a check for /done because we want to be able to access webform submit success page which uses that in the url.
     if ($href != $current_path && strpos($current_path, 'sitesearch') === false && strpos($current_path, '.php') === false && substr($current_path, -5) !== '/done') {
         drupal_not_found();
         drupal_exit();
     }
 
+    // Redirect 403 to drupal_access_denied().
+    $href_access = empty($menu_item['access']);
+
+    if ($href_access) {
```
taeram: From context, I *think* this is checking if the page has 403'd. If so, this variable needs to be more explicit, and be named something like `$is_403_forbidden`, as `$href_access` doesn't tell us much.

sites/all/themes/ucgy_theme/templates/pages/page--page-access-denied.tpl.php
```diff
@@ -0,0 +1,31 @@
...
+  include(path_to_theme() . '/templates/includes/header.inc');
+  drupal_add_http_header('Status', '403 Forbidden');
+?>
+<main id="content">
+  <?php
+    // Don't need the defaults actions link
+    unset($page['content']['system_main']['actions']);
+
+    // Display any Drupal status messages
+    $messages;
```
taeram: Should this have a `print` before it?

sites/all/themes/ucgy_theme/templates/pages/page--page-not-found.tpl.php
```diff
@@ -123,18 +126,16 @@
...
-  <?php $messages; ?>
   <?php
-  /**don't need the defaults actions link ***/
-  unset($page['content']['system_main']['actions']);
-  // content is not needed. Also, this was printing "page /x not found" which is not desired behavior. Removing: print render($page['content']);
+    // Don't need the defaults actions link
+    unset($page['content']['system_main']['actions']);
+
+    // Display any Drupal status messages
+    $messages;
```
taeram: Same here re: `print`

sites/all/themes/ucgy_theme/includes/preprocess.inc
dpfitzsi: Fixed this one. I will pass the TPL comments back to Terry.

sites/all/themes/ucgy_theme/templates/pages/page--page-access-denied.tpl.php
TDRoden: Ah-yup, good catch. Never was provided for the original 404 page...

sites/all/themes/ucgy_theme/templates/pages/page--page-not-found.tpl.php
TDRoden: ...And I just copied the general structure for the 403 page.

PR #842, opened 2017-09-06
WSO-386 - Update Permissions

SandyVallee:
Jesse please look at sh script I added one line but it may not be necessary?


SandyVallee:
just look a the files changed

private/migrations/2017-09-05_wso-386.sh
```diff
@@ -0,0 +1,16 @@
...
+
+set -e -u
+
+if [ -z "$PANTHEON_DEPLOY_ENV" ]; then
+  DRUSH_CMD="drush"
+else
+  DRUSH_CMD="terminus remote:drush $PANTHEON_DEPLOY_ENV -- "
+fi
+# enable the feature
+$DRUSH_CMD en ucgy_core_roles --yes
```
SandyVallee: Not sure if this is necessary? locally it gave me an error the first time I tried a fr so I added this in.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_roles/ucgy_core_roles.features.user_role.inc
```diff
@@ -40,6 +40,12 @@ function ucgy_core_roles_user_default_roles() {
     'weight' => 9,
   );
 
+  // Exported role: Read Only.
+  $roles['Read Only'] = array(
```
SandyVallee: new role that they wanted featurized to go up

private/migrations/2017-09-05_wso-386.sh
dpfitzsi: I think the assumption is that the module (feature) would already be enabled, but this doesn't hurt to have this in to be absolutely sure that it is enabled. If the database returns that the module is already enabled it will just continue the script.

PR #848, opened 2017-09-13
WSO-420 - Update Table Form Permissions

SandyVallee:


sites/all/modules/ucalgary/ucgy_core/ucgy_core_variable_table_form/ucgy_core_variable_table_form.module
```diff
@@ -39,7 +39,7 @@ function ucgy_core_variable_table_form_access_callback() {
   //only grant access of role is Content Administrator & ownership group is "Graduate"
   //user 1 should always have access to it
   if((user_is_logged_in() && in_array('Graduate', $group_name) && in_array('Content Administrator', $user->roles))
-  || $user->uid == 1){
+  || $user->uid == 1 || in_array('Support', $user->roles)){
```
dpfitzsi: Although this code is correct, I would add the Support check right after the check for Content Administrator. Visually I think this would look better.

SandyVallee: Ok so I put it all on one line.  I left the Support checkout outside the brackets because Content Admin has to be also apart of the Graduate group so its inside the  && condition, but they wanted user 1 and Support to not have to be apart of a content group to access, so that's why its outside.  Does that make more sense or is there a better way?

PR #852, opened 2017-09-15
WSO-431 - Resolve errors in ucgy_gbl_countries template

dpfitzsi:
Errors were caused by Views printing the fields regardless of whether they had actual data in them. This check for empty now makes sure that the fields actually have data before attempting to print.


sites/all/modules/ucalgary/ucgy_gbl/ucgy_gbl_countries/templates/views-view--international-admission-requirements--block.tpl.php
```diff
@@ -28,1818 +28,1959 @@
...
-                                                            </div>
-                                                        </div>
-                                                    </ul>
-                                                </div>
-                                            </li>
-
-                                        <?php }; // End Foreach ?>
-
-                                    </ul>
+<?php print render($title_prefix); ?><?php if ($title): ?><?php print $title; ?><?php endif; ?><?php print render($title_suffix); ?><?php if ($header): ?>
```
taeram: This should be split onto multiple lines as it look like PHPStorm reformatted it oddly.

PR #861, opened 2017-09-18
WSO-433 - Resolve Errors - ucgy_gbl_supervisor

SandyVallee:


SandyVallee:
Ok I put ti back in because it was thrown other errors now.  Seems to resolve these errors.  Not sure if it needs the & in front? Seems to remove errors as is on my local branch, not sure how it interacts with other error messages?

dpfitzsi:
for $form_id you don't want the & in front. This should resolve the errors that you were seeing.

PR #862, opened 2017-09-18
WSO-421 - Update Local Users on Non-Live Pantheon Environments

taeram:
This PR updates roles for our test users, as well as adding in the "Arts" Content Ownership Group to each user.

Once deployed, this script will also be called by our `pantheon-import-database` script to create test users on our local development environments.


SandyVallee:
so what will be the usernames and passwords for these accounts?  I think you showed me once how to look this up in the command line , but maybe if you can tell me again before you go that would be great.  Also do these test rules apply to all environments or just local?

private/quicksilver/generate_test_accounts/generate_test_accounts.php
```diff
@@ -1,19 +1,158 @@
...
+
+// The password for each test account
+$accountPassword = 'test';
+
+// The account username, roles and ownership groups
+// This list is sorted from least privileges to most privileges
+$accounts = [
+  'readonly' => [
+    'roles' => [
+      'Read Only'
```
dpfitzsi: Nitpick, for consistency add a comma after 'Read Only'

private/quicksilver/generate_test_accounts/generate_test_accounts.php
```diff
@@ -1,19 +1,158 @@
...
+        // Since the field_* tables use a composite primary key, db_insert won't return the $last_insert_id, so we assume the INSERT was successful
+        db_insert($tableName)->fields([
+          'entity_type' => 'user',
+          'bundle' => 'user',
+          'deleted' => 0,
+          'entity_id' => $userId,
+          'revision_id' => $userId,
+          'language' => 'und',
+          'delta' => $fieldDataDelta++,
+          'field_user_ownership_group_target_id' => $ownershipGroupId,
```
dpfitzsi: Comment: Should this db_insert be able to handle multiple terms?

taeram: Talked offline, leaving as is as it turns out db_insert will only handle a single row at a time

PR #866, opened 2017-09-22
WSo-182 - Nav is double highlighting sites with a secondary overview

SandyVallee:
There maybe a better way to do this with less code.  I worked in this file so that I didn't have to create new hooks from the related js that sits in the contrib section.  Please let me know if you have a cleaner solution or want some of this code rewritten to clean up the file all together?


sites/all/themes/ucgy_theme/includes/theme.inc
```diff
@@ -13,209 +13,217 @@ function ucgy_theme_menu_tree($variables)
...
-      $link = l($element['#title'], $element['#href'], array('html' => TRUE,));
-      // If the link has an active class re-define the current variable to add to the li class
-      if (strpos($link,'class="active') !== FALSE){
-        $current = ' current';
+        $isOverviewLink = FALSE;
+
+        // loop through submenu items to see if more then one are set to active
+        foreach ($element['#below'] as $menuItem) {
+
+          //if sibling menu item has the overview class
```
dpfitzsi: need to add a space at the beginning of comment

SandyVallee: ok done

PR #870, opened 2017-09-29
WSO-468 - Update Site Admin permissions

SandyVallee:
permissions match permissions 3 according to exported csv files of permissions


private/migrations/2017-10-06_wso-468.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+PANTHEON_DEPLOY_ENV=$1
+set -e -u
+if [ -z "$PANTHEON_DEPLOY_ENV" ]; then
+ DRUSH_CMD="drush"
+else
+ DRUSH_CMD="terminus remote:drush $PANTHEON_DEPLOY_ENV -- "
+fi
+# WSO-276- enable new module
+$DRUSH_CMD features-revert-all --yes#!/usr/bin/env bash
```
taeram: As discussed offline, this should not be a `features-revert-all` and should instead be changed to a `fr <module name>`

sites/all/modules/ucalgary/ucgy_core/ucgy_core_roles/ucgy_core_roles.info
```diff
@@ -3,7 +3,70 @@ description = Defines the core roles and permissions used in the sites. This als
 core = 7.x
 package = University of Calgary (Required)
 version = 7.x-1.2
+dependencies[] = admin_menu
```
taeram: As before, the list of dependencies should be trimmed down to just `ucgy_core`, as this module doesn't really need to have every other module in the system as a dependency.

PR #876, opened 2017-10-05
WSO-442- Hero CTA text change

SandyVallee:


SandyVallee:
Take a look. I copied over the 2 feature files and the info file and its only listing these as changes. Maybe I need to test in another environment?

taeram:
Keeping for historical purposes

PR #878, opened 2017-10-10
WSO-498 - Update Future Students

SandyVallee:
is it ok to add a return like a I did?  It works just didn't know if it was good practice or not?


sites/all/themes/ucgy_theme/templates/includes/breadcrumbs.inc
```diff
@@ -1,33 +1,39 @@
 <?php if ($logged_in && $is_admin) { ?><!-- Include file: <?php echo str_replace('-', '=', __FILE__); ?> --><?php } ?>
 <?php // TODO: We should move the below functionality into ucgy_fs ?>
-<?php if (preg_match("/^future-students/", request_path()) && preg_match("/\bgraduate\b/i", request_path())): ?>
-<div id="breadcrumbs" class="row">
+
+<?php
+if (request_path() === 'future-students') {  ?>
+  <div id="breadcrumbs" class="row" style="display:none;"></div>
+<?php return; }
```
taeram: Since everything below is wrapped in if/else's, I think you should be able to remove the `return` withtout affecting the operation of the template.

PR #880, opened 2017-10-11
WSO-458 - Validation module

dpfitzsi:


drush/ucgy_migrations/scripts/1508169891.php
```diff
@@ -0,0 +1,5 @@
+<?php 
+// add your migration step below this line 
+
+// Enable ucgy_core_validation module
+exec("drush en ucgy_core_validation -y");
```
taeram: As mentioned in Hangouts, we should use `system("<command goes here>") || exit(1);" so that we can catch whether the command is successful or not.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_validation/ucgy_core_validation.info
```diff
@@ -0,0 +1,6 @@
+name = University of Calgary Validation
```
taeram: Looking at other modules, I think our standard name for our custom modules is `UCalgary <x>`, making this `UCalgary Validation`

sites/all/modules/ucalgary/ucgy_core/ucgy_core_validation/ucgy_core_validation.module
```diff
@@ -0,0 +1,117 @@
...
+function ucgy_core_validation_form_alter(&$form, &$form_state, $form_id) {
+  // Create form validation callback function
+  $form['#validate'][] = 'ucgy_core_validation_global_form_validate';
+}
+
+// Global validation callback function for drupal forms
+function ucgy_core_validation_global_form_validate($form, &$form_state) {
+
+  // Var used to flag errors
+  $errorFlag = FALSE;
```
taeram: It doesn't look like this variable is used until it's definition below, and could be removed here.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_validation/ucgy_core_validation.module
```diff
@@ -0,0 +1,117 @@
...
+
+            // Check every result record in array
+            foreach ($result as $resultRecord) {
+
+              // Condition used to check if https already exist and if not check for http - based on string position
+              if (strpos($resultRecord, 'https') === FALSE) {
+                if (strpos($resultRecord, 'http') !== FALSE) {
+
+                  // Flag error
+                  $errorFlag = TRUE;
```
taeram: Instead of setting `$errorFlag` and `$errorFound` and then immediately using them on line 106, I think it'd be better to turf these two variables, and simply call `form_set_error()`, removing the `if ($errorFlag...)` conditional that wraps it.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_validation/ucgy_core_validation.module
```diff
@@ -0,0 +1,117 @@
...
+  // Create form validation callback function
+  $form['#validate'][] = 'ucgy_core_validation_global_form_validate';
+}
+
+// Global validation callback function for drupal forms
+function ucgy_core_validation_global_form_validate($form, &$form_state) {
+
+  // Var used to flag errors
+  $errorFlag = FALSE;
+  $errorFound = "";
```
taeram: Same with `$errorFound`, as it doesn't seem to really be used until below.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_validation/ucgy_core_validation.module
```diff
@@ -0,0 +1,117 @@
...
+
+          // Loop through any found html tag results if results have been returned
+          foreach ($results as $result) {
+
+            // Check every result record in array
+            foreach ($result as $resultRecord) {
+
+              // Condition used to check if https already exist and if not check for http - based on string position
+              if (strpos($resultRecord, 'https') === FALSE) {
+                if (strpos($resultRecord, 'http') !== FALSE) {
```
taeram: These two if statements could be combined into a single if with an `&&` separating the conditions.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_validation/ucgy_core_validation.module
```diff
@@ -0,0 +1,117 @@
...
+
+// Global validation callback function for drupal forms
+function ucgy_core_validation_global_form_validate($form, &$form_state) {
+
+  // Var used to flag errors
+  $errorFlag = FALSE;
+  $errorFound = "";
+
+  foreach ($form_state['values'] as $field => $value) {
+    if (is_array($value['und'])) {
```
taeram: Since nothing in this foreach runs if this value isn't set, I think it'd make the code more readable to simply return if this is false: 
```php
if (!is_array($value['und'])) {
    continue;
}

//.. rest of foreach goes here
```

sites/all/modules/ucalgary/ucgy_core/ucgy_core_validation/ucgy_core_validation.module
```diff
@@ -0,0 +1,117 @@
...
+        $errors = [];
+        // Validates node pages using textfields or long text fields
+        if (isset($value['und'][0]['value'])) {
+          if (preg_match_all('/<' . $htmlTag . '[^>]+>/i', $value['und'][0]['value'], $results)) {
+            if (isset($results[0])) {
+              // Add field name for validation
+              $field_name = "$field][und][0][value";
+
+              // Create sub array and set to only use completed array and then merge with main array
+              if (!in_array($field_name, $errors1)) {
```
taeram: As discussed offline, if you can find a way to make this step more readable, that would be great.

drush/ucgy_migrations/scripts/1508169891.php
```diff
@@ -0,0 +1,5 @@
+<?php 
+// add your migration step below this line 
+
+// Enable ucgy_core_validation module
+system("drush en ucgy_core_validation -y");
```
taeram: Should use the new migration helper function.

PR #884, opened 2017-10-16
WSO-530-Update-SimpleSAML-Code

SandyVallee:
This makes me nervous because we can't test it locally, and similar changes in wso-529 were not working.  Please look at these 2 branches together as they are linked.


SandyVallee:
combined with dan into wso-529

PR #891, opened 2017-10-19
WSO-522 implement an image compression module

dpfitzsi:


drush/ucgy_migrations/scripts/1508450925.php
```diff
@@ -0,0 +1,8 @@
+<?php 
+// add your migration step below this line 
+
+// enable ImageInfo Cache module
+system("drush en imageinfo_cache -y");
```
taeram: Could you add a `|| exit(1);` to the end of the system calls, so that if the command fails, we will exit the script with an error state?

drush/ucgy_migrations/scripts/1508450925.php
```diff
@@ -0,0 +1,8 @@
+<?php 
+// add your migration step below this line 
+
+// enable ImageInfo Cache module
+system("drush en imageinfo_cache -y") || exit(1);
+
+// set image styles for inuse image fields
+system("drush image-generate all inuse") || exit(1);
```
taeram: Should use the new ucgy_ helper function

drush/ucgy_migrations/scripts/1508450925.php
```diff
@@ -0,0 +1,8 @@
+<?php
```
taeram: Talking to Nicole, the image-generate step in this migration takes quite some time to run (it's already > 20 minutes on Staging).

As such, we should rename this migration file so its filename (timestamp) is absolutely the last thing that will run on deploy.

We might even want to do a `drush updatedb` and a `drush cc all` before the `image-generate` step so that image-generate can run as long as it has to, and our deploy will be essentially finished before it starts.

PR #903, opened 2017-10-25
WSO-544 Removed testing-ucalgary-future-students-se-views-page views display

dpfitzsi:


taeram:
Closed as this view was deemed in use, and should not be removed.

PR #914, opened 2017-11-24
WSO-652 - Global: Tabbed Content - First Tab displays but not full card

taeram:


sites/all/modules/ucalgary/ucgy_core/ucgy_core_helpers.inc
```diff
@@ -46,7 +46,8 @@ function ucgy_fc_item_is_first(\FieldCollectionItemEntity $field_collection_item
   if (isset($field_collection_item)) {
     $host = $field_collection_item->hostEntity();
     if (isset($host->$fc_obj)) {
-      foreach ($host->$fc_obj[LANGUAGE_NONE] as $key => $item) {
+      $items = $host->$fc_obj;
```
taeram: PHP doesn't allow us to iterate over "$x->$y['foo']". It will just return a NULL.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_helpers.inc
```diff
@@ -46,7 +46,8 @@ function ucgy_fc_item_is_first(\FieldCollectionItemEntity $field_collection_item
...
+      foreach ($items[LANGUAGE_NONE] as $key => $item) {
```
taeram: PHP doesn't allow us to iterate over `$x->$y['foo']`, and instead will just return `NULL`.

PR #921, opened 2017-12-08
WSO-703 - Compile SCSS files to CSS and minify JavaScript on deploy to Pantheon

taeram:
I've split this PR into two separate commits. The first commit is just removing all of the compiled CSS and CSS.map files, so I would recommend only reviewing the [second commit](https://github.com/ucalgary/www.ucalgary/pull/921/commits/033378ba66258dc8363eaa459748851623f875fc) as that's where the actual meat of the changes are.


sites/all/modules/ucalgary/ucgy_gbl/js/javascript.js
```diff
@@ -0,0 +1,21 @@
+(function ($) {
```
dpfitzsi: Why was this file placed in the ucgy_gbl module? It appears to be attached to the ucgy_page_builder module, shouldn't the code be placed there?

PR #922, opened 2017-12-08
WSO-630 - Unpublish options not appearing on Edit page

taeram:


sites/all/modules/ucalgary/ucgy_core/ucgy_core_roles/ucgy_core_roles.module
```diff
@@ -419,23 +417,18 @@ function ucgy_core_roles_contextual_links_view_alter(&$element, $items) {
             ->execute()
             ->fetchAssoc();
```
christophergraha1: Not sure about this query, it would be better to do a where and pull the results you need vs querying multiple times over a foreach.

dpfitzsi:
Conflict in ucgy_core_roles.module. Can this be corrected first?

PR #931, opened 2018-01-31
WSO-807 - Disable Section Path filter for non www.ucalgary sites

taeram:
On non-www.ucalgary.ca sites (e.g. oval.ucalgary.ca, (faculty).ucalgary.ca), we want to allow the pre-footer  / localized footer / section information to be added to the home (front) page.

This PR adds a new variable called `ucgy_section_path_filter_enabled` which if enabled, filters out additional items from the Section Path dropdown.

Also in this PR are some isset() cleanups, as well as coding standards application and template tidying.

Best viewed [with whitespace disabled](https://github.com/ucalgary/www.ucalgary/pull/931/files?w=1)


sites/all/modules/ucalgary/ucgy_core/ucgy_core_section/ucgy_core_section.module
```diff
@@ -163,117 +155,93 @@ function ucgy_core_section_field_widget_info()
...
-            unset($filtered_paths[$key]);
-        }
+function ucgy_core_section_field_widget_form(&$form, &$form_state, $field, $instance, $langcode, $items, $delta, $element) {
+  $is_ucgy_filter_enabled = variable_get('ucgy_section_path_filter_enabled', FALSE);
+
+  // Get the machine names for all menus
+  $menus = array_keys(menu_get_menus());
+
+  // Exclude some menus
+  $excluded_menus = [
```
dpfitzsi: May be want to put this into a constant. We also need to call this during the page builder wizard so having this list defined would be a good idea.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_section/ucgy_core_section.module
```diff
@@ -282,62 +250,78 @@ function ucgy_core_section_preprocess_node(&$vars)
...
+    foreach ($path_parts as $i => $part) {
+      if ($i === 0) {
+        $previous_part = '';
+      }  else {
+        $previous_part = $path_parts[$i - 1] . '/';
+      }
+      $paths[] = $previous_part . $part;
+    }
+  } else {
+    $paths = [$path];
```
dpfitzsi: If the substr_count fails why would the path then be an array?

sites/all/modules/ucalgary/ucgy_core/ucgy_core_section/ucgy_core_section.module
taeram: Good idea, will do.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_section/ucgy_core_section.module
taeram: @dpfitzsi it's a minor change here so we don't have to wrap our `->condition()` below in an if/else to test if `$paths` is an array or not, and adjust the condition accordingly.

We could do the if/else below to simplify the `->condition` as well as remove the `->orderBy` and `->range` if there's just a single path. Probably a good idea, performance-wise.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_section/ucgy_core_section.module
```diff
@@ -35,49 +44,47 @@ function ucgy_core_section_views_api()
...
+      'template' => 'node--section-information--contact-notes',
+    ],
+    'node__section_information__contact_map' => [
+      'variables' => [],
+      'path' => $module_path . '/templates',
+      'base hook' => 'node',
+      'template' => 'node--section-information--contact-map',
+    ],
+  ];
 }
```
christophergraha1: I See that $path and $module_path are the same, this should prob be a const created at the top of the file and used within functions where needed.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_section/ucgy_core_section.module
```diff
@@ -163,117 +165,86 @@ function ucgy_core_section_field_widget_info()
...
+  if ($vars['view_mode'] === 'contact_details') {
+    $vars['theme_hook_suggestions'][] = 'node__' . $vars['type'] . '__contact_details';
+  }
+  if ($vars['view_mode'] === 'contact_notes') {
+    $vars['theme_hook_suggestions'][] = 'node__' . $vars['type'] . '__contact_notes';
+  }
+  if ($vars['view_mode'] === 'contact_map') {
+    $vars['theme_hook_suggestions'][] = 'node__' . $vars['type'] . '__contact_map';
+  }
 }
```
christophergraha1: If these if statements should not be defined as else if, we should have comments explaining why, or else these should be else ifs.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_section/ucgy_core_section.module
```diff
@@ -489,18 +472,17 @@ function ga_details_form_validate($form, &$form_state) {
...
-    //  if we're on the user landing page and the user is a FGS Supervisor, add a class
-    if (in_array("page-user", $variables['classes_array']) && in_array(UCGY_ROLE_FUTURE_GRAD_STUDENTS_SUPERVISOR, $user->roles)) {
-      $variables['classes_array'][] = 'fgs-supervisor';
-    }
-
+  if (arg(0) === "admin") {
+    // reference your current admin theme
+    $theme_path = drupal_get_path('theme', 'seven');
+    // reference your own stylesheet
+    drupal_add_css(drupal_get_path('module', 'ucgy_core_section') . '/css/form.css', ['weight' => CSS_THEME]);
```
christophergraha1: Could use path const here, if one is created.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_section/ucgy_core_section.module
taeram: Switched them all to if / elseif, because there's no way the same variable could ever equal more than one thing at a time, so an if/elseif is appropriate.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_section/ucgy_core_section.module
taeram: Replaced this with `$path` that's passed in the function params, as it contains the path to this module.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_section/ucgy_core_section.module
taeram: Left as-is for now. Defining this as a constant would be a larger standards change we'd want to look at, as we would want to maintain the same standards across all of our modules.

sites/all/modules/ucalgary/ucgy_core/ucgy_core_section/ucgy_core_section.module
```diff
@@ -486,21 +437,16 @@ function ga_details_form_validate($form, &$form_state) {
...
-    }
-
-    // get the user profile
-    global $user;
-    //  if we're on the user landing page and the user is a FGS Supervisor, add a class
-    if (in_array("page-user", $variables['classes_array']) && in_array(UCGY_ROLE_FUTURE_GRAD_STUDENTS_SUPERVISOR, $user->roles)) {
-      $variables['classes_array'][] = 'fgs-supervisor';
-    }
-
+  if (arg(0) === "admin") {
```
dpfitzsi: Would be better to apply the path using the function path_is_admin(): https://api.drupal.org/api/drupal/includes%21path.inc/function/path_is_admin/7.x

taeram: That is a handy function. Thanks!

taeram: That's a handy function, thanks!

PR #937, opened 2018-02-09
WSO 814 404 JS error

TDRoden:


sites/all/themes/ucgy_theme/resources/js/404.js
```diff
@@ -0,0 +1,61 @@
+(function($){
+
+  $(document).ready(function() {
+
+    // Update the user's options list
+    if (document.referrer!='') {
```
taeram: Should use `!==` to enforce strict type checking.

(I realize these comments are on code that wasn't written in this ticket, but I would still like to get them updated to move our code forward in terms of overall quality)

sites/all/themes/ucgy_theme/resources/js/404.js
```diff
@@ -0,0 +1,61 @@
...
+      $('#options').prepend('<li>Returning to your <a id="back">previous</a> website location.</li>');
+      $('#back').css('cursor', 'pointer').click(function() {
+        parent.history.back();
+      });
+    }
+
+    // Test for audio support and modify accordingly
+    var audioTagSupport = !!(document.createElement('audio').canPlayType);
```
taeram: The use of `!!` is a little "lazy". I'd prefer to use a more explicit check.

In this case, this could be changed to: 
```js
var audioTagSupport = typeof document.createElement('audio').canPlayType === "function";
```

PR #941, opened 2018-02-15
WSO-812 - Unable to create new entitites

taeram:


sites/all/modules/ucalgary/teams/teams_entity/teams_entity.module
```diff
@@ -49,15 +49,19 @@ function teams_entity_access($op, $entity = NULL, $account = NULL) {
   }
 
   // Handle "Add X" links.
-  if (is_string($entity) && $op === "create" && user_access('eck add entity types', $account)) {
-    return TRUE;
+  if (is_string($entity) && $op === "create") {
+    return (
+      user_access("eck add $entity $entity entities", $account) ||
```
dpfitzsi: This should work, but I do want to remove any references to ECK module from this module if possible. I would rather it just be designed to work with generic entities. OR we could wrap these specific permissions in the future with a "if (module_exists('eck')) {"

PR #949, opened 2018-02-27
WSO-813 - Teams override for node authors

taeram:
This PR adds the "teams_override" module, which allows a user to update a node if:
* they are the author of that node
* they have "edit any `<node type>` content" permissions

Teams Superadmins will see a "Teams Override" tab at the bottom of any `node/%node/edit` page, where they can enable/disable the override, and set the author of the node.

Finally, "FGS - Supervisor" role has been removed from the system entirely, and replaced with a database check. An FGS supervisor is now "a user who is the author of a node that has a node type of `gbl_supervisor`". 

Supervisor pages can be added for a user by Support:
* Support user logs in
* Edits a user
* Checks off the "Add a Supervisor page for this user" checkbox
* Saves the user
* The supervisor page is created (as before) in the background

To remove the supervisor page form, a Support deletes the user's support page from the site.


drush/ucgy_migrations/scripts/1519681286.php
```diff
@@ -0,0 +1,69 @@
...
+//    // Run a shell command
+//    ucgy_migrations_shell_command("ls -1 /tmp/");
+//
+//    // Feature revert a module
+//    ucgy_migrations_shell_command("drush fr ucgy_my_module --yes");
+//
+//    // Run an SQL Query
+//    db_query("UPDATE users SET name = 'bill1234' WHERE name = 'bill'");
+
+$fgs_supervisor_team_id = db_query("SELECT team_id FROM teams_teams WHERE name = 'FGS - Supervisor'")->fetchField();
```
dpfitzsi: Until we clean up these migrations scripts this can be a security threat (SQL Injection). Either we can switch these db_query select statements or switch to using placeholders inside the db_query. For the latter:

`db_query("SELECT team_id FROM teams_teams WHERE name = :name", [':name' => 'FGS - Supervisor'])->fetchField();`

This would apply to each of the db_query statement below

drush/ucgy_migrations/scripts/1519752555.php
```diff
@@ -0,0 +1,18 @@
...
+//    // Feature revert a module
+//    ucgy_migrations_shell_command("drush fr ucgy_my_module --yes");
+//
+//    // Run an SQL Query
+//    db_query("UPDATE users SET name = 'bill1234' WHERE name = 'bill'");
+
+// Delete the unused edit supervisor role
+//  ** I'm removing this from the database manually, as using the drush role-delete command causes
+//     a bunch of "Click here to edit <name of a piece of content>" links to show up on the /user/login page
+$fgs_supervisor_rid = db_query("SELECT rid FROM role WHERE name = 'FGS - Supervisor'")->fetchField();
```
dpfitzsi: The same query construction should apply here

sites/all/modules/ucalgary/teams/teams_override/includes/admin.inc
```diff
@@ -0,0 +1,156 @@
+<?php
```
dpfitzsi: I think this implementation, while it works isn't a very good way of doing this. This would better implemented by creating an API hook for overriding the forms that we're defining in the teams module. The original form would then contain a drupal_alter hook where we could pass in new paramaters. This way a lot of the duplicated code would be reduced. 

Alternatively, should this be its own module in the first place? If we feel the necessity for doing overrides just for nodes should this be a configurable admin form inside the node module? 

Also is this a such as specific use case that only we need this functionality. If so, should this be removed from the teams ecosystem and instead placed in ucalgary/ucgy_teams/teams_override.

sites/all/modules/ucalgary/teams/teams_override/teams_override.info
```diff
@@ -0,0 +1,7 @@
+name = Teams Override
+description = Override teams access for specific nodes.
+core = 7.x
+package = Teams
+
+dependencies[] = teams
+dependencies[] = node
```
dpfitzsi: In this case we would want to also have a dependency on teams_nodes

sites/all/modules/ucalgary/ucgy_gbl/ucgy_gbl_supervisor/ucgy_gbl_supervisor.module
```diff
@@ -289,210 +299,170 @@ function ucgy_gbl_supervisor_js_alter(&$javascript) {
...
 
-    if (isset($pass) && !empty($pass)) {
-      db_insert("ucgy_core_protected_pages")
-        ->fields(["password" => sha1(trim(check_plain($pass))), "path" => check_plain("node/" . $node->nid . "/edit"),])
-        ->execute();
-    }
+  // Add the user to the FGS Supervisor team
+  $fgs_supervisor_team_id = db_select('teams_teams', 'tt')
+    ->fields('tt', ['team_id'])
+    ->condition('name', UCGY_GBL_SUPERVISOR_FUTURE_GRAD_STUDENTS_SUPERVISOR_TEAM)
```
dpfitzsi: `->condition('name', UCGY_GBL_SUPERVISOR_FUTURE_GRAD_STUDENTS_SUPERVISOR_TEAM, '=')`

PR #951, opened 2018-03-05
OP-120-Fix social media image when sharing ucalgary pantheon urls

mgbauman:
Fixed issue with the social media images not displaying correctly by adding token to default Open Graph image settings.


drush/ucgy_migrations/scripts/1520014850.php
```diff
@@ -0,0 +1,21 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: mgbauman
+ * Date: 3/2/2018
+ * Time: 11:19 AM
+ */
```
taeram: Should remove this PHPStorm comment, as it doesn't really add any value, but since this is a migration (which gets deleted at some point in the future), don't worry about it for this branch.

PR #958, opened 2018-03-23
OP-86-Error logs 4

mgbauman:
Removed indications of errors with logs turned on.


sites/all/modules/ucalgary/ucgy_core/ucgy_core_layout/ucgy_core_layout.module
```diff
@@ -545,8 +545,10 @@ function _get_fpp__panels_pid($fpid) {
         //load the fpp data from subtype
         $fpp_info = _ucgy_core_get_fpp_from_subtype($subtype);
         //check for fpid and then return the subtype
-        if ($fpp_info->fpid == $fpid) {
-          return ($pid);
+        if(is_object($fpp_info)) {
```
taeram: This should be changed to `if (isset($fpp_info->fpid)) {` as what we really care about is if the `fpid` parameter exists in the `$fpp_info` object.

PR #962, opened 2018-04-02
OP-149-Fix Cloning

mgbauman:
Added new module named ucgy_teams_override. Contains function to properly set teams for cloned pages and their bricks.


sites/all/modules/ucalgary/ucgy_teams_override/ucgy_teams_override.info
```diff
@@ -0,0 +1,8 @@
+name = Teams Override
```
dpfitzsi: I think the name space should be something like ucgy_teams/teams_override. This way your work on permissions can follow a similar pattern

sites/all/modules/ucalgary/ucgy_teams_override/ucgy_teams_override.info
```diff
@@ -0,0 +1,8 @@
...
+description = Overrides the teams module to allow for additional functionality specific to the University of Calgary
+core = 7.x
+package = Teams
+
```
dpfitzsi: Remove extra space

drush/ucgy_migrations/scripts/1522691003.php
```diff
@@ -0,0 +1,3 @@
+<?php
+
+ucgy_migrations_shell_command("drush en ucgy_teams_override");
```
dpfitzsi: This would need to update if you change the module name again...

drush/ucgy_migrations/scripts/1522691003.php
```diff
@@ -0,0 +1,3 @@
+<?php
+
+ucgy_migrations_shell_command("drush en teams_override");
```
dpfitzsi: command will need to end in a -y so it processes and doesn't wait for user input

`ucgy_migrations_shell_command("drush en teams_override -y"); `

PR #964, opened 2018-04-02
OP-142-File Upload Teams Bug

mgbauman:
Added migration file to move all files from the old groups folder to the new teams folder.


drush/ucgy_migrations/scripts/1522702354.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+$old_path = DRUPAL_ROOT."/sites/default/files/groups/";
+$new_path = DRUPAL_ROOT."/sites/default/files/teams/";
+
+ucgy_migrations_shell_command("cp $new_path'marketing/University-Fact-Sheet.pdf' $old_path'marketing/'");
```
dpfitzsi: Was this a link used for debugging. It appears that this file would be picked up based on directories in the array on line 8.

PR #967, opened 2018-04-19
OP-179-Database change for teams_entity table

mgbauman:
Changed teams_entity table. Added migration script that grabs team information from the content ownership table.


sites/all/modules/ucalgary/teams/teams_entity/teams_entity.module
```diff
@@ -61,7 +61,7 @@ function teams_entity_access($op, $entity = NULL, $account = NULL) {
       $permission_action = 'add';
     }
     if (isset($entity->id) && user_access('eck ' . $permission_action . ' entity types', $account)) {
-      $entity_teams_access = _teams_entity_find_entities_by_eid($entity->id);
+      $entity_teams_access = _teams_entity_find_entities_by_eid($entity->id,$entity->type);
```
dpfitzsi: Should space the two variables $entity->id, $entity->type

sites/all/modules/ucalgary/teams/teams_entity/teams_entity.install
```diff
@@ -29,6 +29,13 @@ function teams_entity_schema() {
         'not null' => TRUE,
         'default' => 0,
       ],
+      'entity_type' => [
+        'description' => 'The type of the entity in Drupal.',
```
dpfitzsi: sp: The type of entity in Drupal.

sites/all/modules/ucalgary/teams/teams_entity/teams_entity.install
```diff
@@ -55,6 +62,20 @@ function teams_entity_schema() {
 
 }
 
+function teams_entity_update_7001() {
+
+  if (!db_field_exists('teams_entity', 'entity_type')) {
+    $field = [
+      'description' => 'The type of the entity in Drupal.',
```
dpfitzsi: sp: The type of entity in Drupal.

drush/ucgy_migrations/scripts/1523903862.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+ucgy_migrations_shell_command("drush updb -y");
+
+db_truncate('teams_entity')->execute();
```
dpfitzsi: Will this potentially destroy new data?

mgbauman: My objective was to clear the teams_entity table so that it can be repopulated with information from the Content Ownership table. I believe this achieves that. Is there a different method I should be using?

dpfitzsi: This may work the way we want, but for safety you could also use a db_rename_table to rename the existing table to something else to preserve it until we want to delete it later and before you run the new copy from the Content Ownership table you could recreate the teams_entity table using this method: https://drupal.stackexchange.com/questions/145657/create-new-table-in-existing-module-without-affecting-existing-table-data (second answer calling schema)

PR #979, opened 2018-05-14
OP-197 Gap between Banner and Section Information Prefooter

mgbauman:
Set margin of the banner block to 0 when corners are rounded. Changed z-index of pre footer so that even if a brick would cover it (With rounded corners for example) the pre-footer will appear on top.


taeram:
Closing this PR, and deleting this branch as it's no longer required.

PR #984, opened 2018-06-13
OP-167 Bulk Update Confirmation Bug

nindusaini:
additional code to prevent error message even after success


sites/all/modules/ucalgary/teams/includes/views_bulk_operations.inc
```diff
@@ -97,13 +97,26 @@ function teams_assign_content_arguments(&$entity, $context) {
     }
   }
 
-  // Clear existing data for this value
-  $teams_delete_teams = db_delete($table)
-    ->condition($column, $field, '=')
-    ->execute();
+  /***
```
mgbauman: I think 2 * is enough.

sites/all/modules/ucalgary/teams/includes/views_bulk_operations.inc
```diff
@@ -97,13 +97,26 @@ function teams_assign_content_arguments(&$entity, $context) {
...
-  // Clear existing data for this value
-  $teams_delete_teams = db_delete($table)
-    ->condition($column, $field, '=')
-    ->execute();
+  /**
+   * We're getting error message 'Error removing team' if team data for that particular field is not in the teams.entity_table_name tables i.e a new entry
+   * So let's place a check that this message does not show for a new entry
+   */
 
-  if (empty($teams_delete_teams)) {
```
taeram: If there were no rows deleted, `$teams_delete_teams` should equal `0`. However, if there was an error deleting the teams, `$teams_delete_teams` should equal `NULL`, which is what this condition should be testing for.

I think this code block could simply be changed to the following:
```php
  // Clear existing data for this value
  $teams_delete_teams = db_delete($table)
    ->condition($column, $field, '=')
    ->execute();
  if ($teams_delete_teams === NULL) {
    drupal_set_message(t('Error removing team'), 'error');
  }
```

